Issue #9, opened 2016-10-12
Write a utility that acts as wrapper to ansible modules


nnDarshan:
Write a utility in tendrl node agent,  that can be used to leverage the functionality provided by ansible-modules .  This utility must act as a wrapper on top of ansible modules.

Corresponding JIRA ticket: https://tendrl.atlassian.net/browse/TEN-26

PR #13, opened 2016-10-13
Added a utility to execute ansible modules.

nnDarshan:
This patch adds a wrapper on top of ansible modules, using
which we can execute the tasks that are done by ansible
module locally.

Issue #9


tendrl/node_agent/ansible_runner/ansible_module_runner.py
```diff
@@ -0,0 +1,73 @@
...
+                       "module %s , with arguments %s. Error:" + \
+                       " %s".format(module_path, arguments, err)
+
+
+class AnsibleRunner(object):
+    """Class that can be used to run ansible modules
+
+    """
+    def __init__(self, module_path, **kwargs):
+        self.executable_module_path = "/tmp/.tendrl_runner" + str(
```
nthomas-redhat: why hard-coding the path here? Either a const or part of some config file?

tendrl/node_agent/flows/execute_command.py
```diff
@@ -0,0 +1,13 @@
...
+
+class ExecuteCommand(object):
+    def __init__(self, api_job):
+        super(ExecuteCommand, self).__init__()
+        self.api_job = api_job
+        self.atom = Command
+
+    def start(self):
+        attributes = self.api_job['attributes']
+        result, err = self.atom().start(attributes)
```
shtripat: Shouldnt we take _raw_input from attributes and pass to start to be specific. I understand inside atom's start() function you might be doing that.

tendrl/node_agent/manager/rpc.py
```diff
@@ -0,0 +1,111 @@
...
+        self.client = etcd.Client(**etcd_kwargs)
+        self.bridge_id = str(uuid.uuid4())
+
+    def _acceptor(self):
+        while True:
+            jobs = self.client.read("/api_job_queue")
+            for job in jobs.children:
+                raw_job = json.loads(job.value.decode('utf-8'))
+                # Pick up the "new" job that is not locked by any other bridge
+                if raw_job['status'] == "new":
```
shtripat: As discussed we should have some mechanism to figure out if the job is actually meant for the node agent and not for the storage bridge here. Just based on status of the job, there could be mix of storage bridge job and node agent jobs being picked here for processing.

tendrl/node_agent/manager/rpc.py
```diff
@@ -0,0 +1,111 @@
...
+                    try:
+                        raw_job['status'] = "processing"
+                        # Generate a request ID for tracking this job
+                        # further by tendrl-api
+                        req_id = str(uuid.uuid4())
+                        raw_job['request_id'] = "%s/flow_%s" % (
+                            self.bridge_id, req_id)
+                        self.client.write(job.key, json.dumps(raw_job))
+                        gevent.sleep(2)
```
shtripat: Need to move this sleep outside for loop ?

tendrl/node_agent/ansible_runner/ansible_module_runner.py
nnDarshan: Will Do that

tendrl/node_agent/flows/execute_command.py
nnDarshan: All the key-value pairs of the attributes dictionary act as Key word args for the runner function. For example, in case of package installation atom the attributes dict will be something like
{
"name" : "gluster",
"state": "installed"
}
So the whole dictionary is needed by atom.

tendrl/node_agent/manager/rpc.py
nnDarshan: yes, Will Do that.

tendrl/node_agent/manager/rpc.py
nnDarshan: Yes, I feel we can use the convention to have 
sds_type : gluster ---> for gluster specific jobs
sds_type: geph     ---> for ceph specific jobs
sds_type: generic ---> for jobs intended to be picked by node agent
while defining the job
Please Provide your thoughts : @shtripat @nthomas-redhat @r0h4n

nnDarshan: Another Option I see is having separate job queues for different jobs namely gluster_job_queue, ceph_job_queue, node_agent_job_queue from where gluster-bridge ceph-bridge and node-agent will pick the jobs respectively

nnDarshan:
@shtripat @nthomas-redhat @r0h4n Please review the pull request. Have addressed previous comments and added unit tests.

tendrl/node_agent/manager/rpc.py
nthomas-redhat: I am inline with filter based on sds_type, rather than having separate queues

tendrl/node_agent/manager/rpc.py
```diff
@@ -23,48 +25,64 @@ def __init__(self):
...
+                    "error": err
+                }
+                return raw_job, True
+        else:
+            return raw_job, False
+
     def _acceptor(self):
         while True:
             jobs = self.client.read("/api_job_queue")
+            gevent.sleep(2)
```
nthomas-redhat: What is the purpose of this sleep?
Why its taken out from the for loop?

nnDarshan: This reading from the job queue is done through greenlets(kind of light-wait threads). These threads have to make way for other threads by sleeping periodically. Else the other threads starve. Here job-queue executing greenlet sleeps for 2 secs and wakes up and checks job queue for new jobs in a continuous loop.
Earlier when the sleep was inside, the sleep was not getting called once all the new jobs in the queue are processed, because of this other greenlets were starved. To fix that it was moved out

nthomas-redhat: fine, got it.

tendrl/node_agent/manager/rpc.py
shtripat: I agree with @nthomas-redhat and that could be easy hack.
Long term I feel having different job queues would be a better option but not for 1.0 release :)

doc/source/installation.rst
```diff
@@ -1,3 +1,10 @@
+===========
```
r0h4n: https://github.com/Tendrl/ceph_bridge/pull/15

Make this install file as the above PR, move the sample job creation parts to some other doc. not here

tendrl/node_agent/ansible_runner/ansible_module_runner.py
```diff
@@ -0,0 +1,78 @@
+import ansible.executor.module_common as module_common
```
r0h4n: Remove this file from the PR, since we just want to merge the framework bits and not the ansible task related bits

tendrl/node_agent/atoms/generic/command.py
```diff
@@ -0,0 +1,20 @@
+from tendrl.node_agent.ansible_runner.ansible_module_runner \
```
r0h4n: Remove this file from the PR, since we just want to merge the framework bits and not the ansible task related bits

tendrl/node_agent/flows/execute_command.py
```diff
@@ -0,0 +1,17 @@
+from tendrl.node_agent.atoms.generic.command import Command
```
r0h4n: Remove this file from the PR, since we just want to merge the framework bits and not the ansible task related bits

tendrl/node_agent/flows/flow_execution_exception.py
```diff
@@ -0,0 +1,4 @@
+class FlowExecutionFailedError(Exception):
```
r0h4n: Remove this file from the PR, since we just want to merge the framework bits and not the ansible task related bits

tendrl/node_agent/manager/__init__.py
```diff
@@ -0,0 +1,3 @@
+
```
r0h4n: why is variable required in the init file

tendrl/node_agent/manager/manager.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self):
+        self._complete = gevent.event.Event()
+
+        self._user_request_thread = EtcdThread(self)
+
+    def stop(self):
+        LOG.info("%s stopping" % self.__class__.__name__)
+        self._user_request_thread.stop()
+
+    def _recover(self):
```
r0h4n: remove this function, seems unecessary

tendrl/node_agent/manager/rpc.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self):
+        etcd_kwargs = {'port': int(config.get("bridge_common", "etcd_port")),
+                       'host': config.get("bridge_common", "etcd_connection")}
+
+        self.client = etcd.Client(**etcd_kwargs)
+        self.bridge_id = str(uuid.uuid4())
+
+    def _process_job(self, raw_job, job_key):
+        # Pick up the "new" job that is not locked by any other bridge
+        if raw_job['status'] == "new" and raw_job["sds_type"] == "generic":
```
r0h4n: I think we need to separate "sds" jobs and "node" jobs, basically have a raw_job["type"] = sds|node
Please communicate this to @anivargi

test-requirements.txt
```diff
@@ -2,12 +2,14 @@
 # of appearance. Changing the order has an impact on the overall integration
 # process, which may cause wedges in the gate later.
 
+mock
 hacking<0.12,>=0.11.0 # Apache-2.0
 pytest
 pytest-cover
+ansible>=2.1.1
```
r0h4n: remove this, make it part of a new PR for the ansible related work (since we need to add more details related to valid shell commands etc)

tendrl/node_agent/flows/flow_execution_exception.py
nnDarshan: I have retained this as a place holder exception that I expect flows to raise if something goes wrong. This is handled in RPC layer.

nnDarshan:
@r0h4n Have trimmed the PR to have only framework related changes

PR #17, opened 2016-10-25
Initial tendrl definitions for node_agent

r0h4n:
- Confirms to tendrl definitions schema 0.1
- Also adds sample config file for node_agent

### DIFF
--- a/etc/tendrl/tendrl.conf.sample
+++ b/etc/tendrl/tendrl.conf.sample
@@ -1,0 +1,17 @@
+[bridge_common]
+
+# Central store etcd port
+etcd_port = 2379
+
+# Central store etcd host/ip
+etcd_connection = 0.0.0.0
+
+# Path to log file
+log_path = /var/log/tendrl/bridge_common.log
+log_level = DEBUG
+
+
+[node_agent]
+# Path to log file
+log_path = /var/log/tendrl/node_agent.log
+log_level = DEBUG
--- a/etc/tendrl_definitions_node_agent.yaml
+++ b/etc/tendrl_definitions_node_agent.yaml
@@ -1,0 +1,92 @@
+---
+tendrl_schema_version: 0.1
+
+valid_objects:
+- tendrl_context
+- cpu
+- network
+- disk
+- memory
+- lvm
+- iptable
+- user
+- group
+- node
+- process
+- proc
+
+object_details:
+  cpu:
+    enabled: True
+    value: "/nodes/$node.id/cpu"
+    attrs:
+      user:
+        type: String
+        help: "Amount of time the CPU was busy executing code in user space"
+      system:
+        type: String
+      wait:
+        type: String
+      nice:
+        type: String
+      swap:
+        type: String
+      interrupt:
+        type: String
+      softirq:
+        type: String
+      steal:
+        type: String
+      idle:
+        type: String
+      active:
+        type: String
+    atoms:
+        set:
+          run: "tendrl.node_agent.atoms.cpu.set"
+          type: "update"
+          pre_run : ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+          post_run: ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+          name: Meaningful name for this atom, should be something that is distinguishable and understandable for notification, debugging, storing and any other similar purposes
+          outputs:
+          - "A map of type {object.attr1 : {required: True}, object.attr2:{required: False}}}} that this will be providing (or could provide) to others, used to correlate and associate the thing/s this atom produces, if it produces anything at all."
+          inputs:
+          - An immutable {object1.attr1 : {value: "cluster1"}, object.attr2:{value: "123?"}}}} dictionary which specifies any initial inputs that should be automatically injected into the atoms scope before the atom execution commences (this allows for providing atom local values that do not need to be provided by other atoms/dependents).
+          version: An immutable version that associates version information with this atom. It can be useful in resuming older versions of atoms. Standard major, minor versioning concepts should apply
+
+  network:
+    enabled: True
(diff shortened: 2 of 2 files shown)


r0h4n:
@nnDarshan Please use this a starting point for your ansible related work. Your ansible wrapper will be invoked by atoms of the objects defined in "tendrl_definitions_node_agent.yaml" in this PR

r0h4n:
@nthomas-redhat  Please review

r0h4n:
@brainfunked Please review, need to merge this soon

etc/tendrl/tendrl.conf.sample
```diff
@@ -0,0 +1,17 @@
+[bridge_common]
```
nnDarshan: should we have separate conf  files for different components like node_agent, and bridges. so that during rpm installation we can copy them to respective locations. Instead of having all the details in sample files of all the conf files.
Just a thought.

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
+---
+tendrl_schema_version: 0.1
+
+valid_objects:
+- tendrl_context
+- cpu
+- network
+- disk
+- memory
+- lvm
```
nnDarshan: when we have disk should we have lvm as well ?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - "A map of type {object.attr1 : {required: True}, object.attr2:{required: False}}}} that this will be providing (or could provide) to others, used to correlate and associate the thing/s this atom produces, if it produces anything at all."
+          inputs:
+          - An immutable {object1.attr1 : {value: "cluster1"}, object.attr2:{value: "123?"}}}} dictionary which specifies any initial inputs that should be automatically injected into the atoms scope before the atom execution commences (this allows for providing atom local values that do not need to be provided by other atoms/dependents).
+          version: An immutable version that associates version information with this atom. It can be useful in resuming older versions of atoms. Standard major, minor versioning concepts should apply
+
+  network:
+    enabled: True
+    value: "/nodes/$node.id/networks/$network.id"
+    attrs:
+      brickdetails:
```
nnDarshan: please remove this

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
...
+        type: String
+        help: Details of the brick
+
+
+flows:
+  CreateCephCluster:
```
nnDarshan: please change this.
I guess this would be something like get_hardware_inventory , which would have cpu/network/disk/memory as atoms
right ?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
+---
+tendrl_schema_version: 0.1
+
+valid_objects:
+- tendrl_context
+- cpu
+- network
+- disk
+- memory
```
nnDarshan: can you please include swap ?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
...
+- lvm
+- iptable
+- user
+- group
+- node
+- process
+- proc
```
nnDarshan: what would this object cater to, can you please elaborate ?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -0,0 +1,92 @@
...
+      softirq:
+        type: String
+      steal:
+        type: String
+      idle:
+        type: String
+      active:
+        type: String
+    atoms:
+        set:
```
nnDarshan: should this be get() as we are getting the cpu info from the system ?

etc/tendrl/tendrl.conf.sample
r0h4n: conf files are not ideally created/copied by rpm installation methods, We prefer if the admin/tendrl install scripts do it. Lets revisit this if we face any nags in future

etc/tendrl_definitions_node_agent.yaml
r0h4n: please add that in one of your patch, this is not a final list

etc/tendrl_definitions_node_agent.yaml
r0h4n: yes, because LVM and its actions work on abstractions above the "disk"

etc/tendrl_definitions_node_agent.yaml
r0h4n: this is to allow listing items in /proc

etc/tendrl_definitions_node_agent.yaml
r0h4n: we dont expose getters via flows/atoms, they are directly fetched from "value" of the object

etc/tendrl_definitions_node_agent.yaml
nnDarshan: sure

etc/tendrl_definitions_node_agent.yaml
r0h4n: it is just a guidance doc, we can enhance/correct it later, the intent here is to exhibit the yaml schema

etc/tendrl_definitions_node_agent.yaml
r0h4n: Create Cluster is a Node agent flow.

also we dont use flows/atoms to "get" objects, we use their "value" directly to read from central store

nnDarshan: okay.

nnDarshan:
Patch looks good to me +1.

PR #20, opened 2016-10-31
Add files to build rpm

TimothyAsir:
Add files: Makefile, node_agent.spec, tendrl-noded.service
and updates setup.py for rpm build feature


TimothyAsir:
@r0h4n, @nthomas-redhat , @ktdreyer, @shripat, @nnDarshan , @anmolbabu , Please review

node_agent.spec
```diff
@@ -0,0 +1,67 @@
+%define pkg_name tendrl-node-agent
+%define pkg_version 0.0.1
+%define pkg_release 1
+
+Name: %{pkg_name}
+Version: %{pkg_version}
+Release: %{pkg_release}%{?dist}
```
ktdreyer: This boilerplate duplication can be simplified:

```
Name: tendrl-node-agent
Version: 0.0.1
Release: 1%{?dist}
```

And then use `%{name}` and `%{version}` macros throughout the packaging, instead of the non-standard `%{pkg_name}` or `%{pkg_version}`

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+BuildArch: noarch
+Summary: Common Module for All Bridges
+Source0: %{pkg_name}-%{pkg_version}.tar.gz
+Group:   Applications/System
```
ktdreyer: No need for `Group:`. I recommend you drop this line.

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+License: LGPL2.1
```
ktdreyer: This `License:` field should be one of the values from https://fedoraproject.org/wiki/Licensing:Main

For example, "License: LGPLv2"

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+Url: https://github.com/Tendrl/node_agent
```
ktdreyer: Url -> URL

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+sphinx-1.0-build doc/source html
+%else
+sphinx-build doc/source html
+%endif
+# remove the sphinx-build leftovers
+rm -rf html/.{doctrees,buildinfo}
+
+%install
+%{__python} setup.py install --single-version-externally-managed -O1 --root=$RPM_BUILD_ROOT --record=INSTALLED_FILES
+install -Dm 0644 tendrl-noded.service $RPM_BUILD_ROOT/usr/lib/systemd/system/tendrl-noded.service
```
ktdreyer: Please use the systemd macro `%{_unitdir}` instead of hard-coding `/usr/lib/systemd/system` here.

(See `rpm --eval %{_unitdir}` to view its value.)

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+%preun
+%systemd_preun tendrl-noded.service
+
+%postun
+%systemd_postun_with_restart tendrl-noded.service
+
+%files -f INSTALLED_FILES
+%doc html README.rst
+%license LICENSE
+%{_usr}/lib/systemd/system/tendrl-noded.service
```
ktdreyer: Please use `%{_unitdir}` here.

setup.py
```diff
@@ -1,13 +1,6 @@
...
-
-def extract_requirements(filename):
-    with open(filename, 'r') as requirements_file:
-        return [x[:-1] for x in requirements_file.readlines()]
-
-install_requires = extract_requirements('requirements.txt')
-
 setup(
     name="tendrl_node_agent",
     version="0.1",
```
ktdreyer: This version "0.1" does not match what is in the RPM packaging above.

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+License: LGPL2.1
+Url: https://github.com/Tendrl/node_agent
+
+Requires: python-etcd >= 0.4.3
+Requires: python-gevent >= 1.0.2
+Requires: python-greenlet >= 0.3.2
+Requires: python-taskflow >= 2.6
+Requires: collectd >= 5.5.1
+Requires: python-jinja2 >= 2.7.2
+Requires: tendrl-bridge-common
```
ktdreyer: Please add `Requires: systemd` as well.

There are no `BuildRequires:` lines here. I think you'll need (at a minimum):

```
BuildRequires: systemd
BuildRequires: python2-devel
BuildRequires: python-sphinx
```

... but you'll need to run this through the "mock" build tool to be certain. Does this package build in mock for EL7? Try rebuilding your SRPM to find out:

```
mock -r epel-7-x86_64 rebuild tendrl-node-agent-0.0.1-1.el7.src.rpm
```
(In fact it would probably be safer if the Makefile used `mock` instead of calling `rpmbuild` directly.)

node_agent.spec
```diff
@@ -0,0 +1,67 @@
...
+%else
+sphinx-build doc/source html
+%endif
+# remove the sphinx-build leftovers
+rm -rf html/.{doctrees,buildinfo}
+
+%install
+%{__python} setup.py install --single-version-externally-managed -O1 --root=$RPM_BUILD_ROOT --record=INSTALLED_FILES
+install -Dm 0644 tendrl-noded.service $RPM_BUILD_ROOT/usr/lib/systemd/system/tendrl-noded.service
+
```
ktdreyer: Please run the test suite during the RPM build to ensure that the tests pass for this specific OS platform and Python version. You'll need to add `BuildRequires: pytest`, and then add a `%check` section like so:

```
%check
py.test -v tendrl/node_agent/tests
```

node_agent.spec
TimothyAsir: Sure

setup.py
TimothyAsir: I will change to 0.0.1

node_agent.spec
TimothyAsir: Sure

TimothyAsir:
Updated the spec file as per review comment. Please review the updated patch
@r0h4n, @ktdreyer, @nthomas-redhat, @shtripat

TimothyAsir:
Updated the spec file as per review comment.
Changed the license to LGPLv2+.
 
Please review the updated patch
@r0h4n, @ktdreyer, @nthomas-redhat, @shtripat

node_agent.spec
```diff
@@ -0,0 +1,65 @@
...
+Requires: collectd
+Requires: python-jinja2
+Requires: tendrl-bridge-common
+Requires: systemd
+
+%description
+Python module for Tendrl node bridge to manage storage node in the sds cluster
+
+%prep
+%setup %{name}-%{version}
```
ktdreyer: Please just use `%setup` here.

Did you try building this package in mock for EPEL 7?

PR #23, opened 2016-11-09
ImportGlusterCluster Flow

r0h4n:


r0h4n:
@nthomas-redhat @nnDarshan @anivargi

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -2,22 +2,34 @@
...
+        help: "Name of the Tendrl managed sds, eg: 'gluster'"
+        type: String
+      sds_version:
+        help: "Version of the Tendrl managed sds, eg: '3.2.1'"
+        type: String
+    enabled: true
+
+  Cpu:
     enabled: True
     value: "/nodes/$node.id/cpu"
```
shtripat: not required

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -2,22 +2,34 @@
...
+        type: String
+      sds_name:
+        help: "Name of the Tendrl managed sds, eg: 'gluster'"
+        type: String
+      sds_version:
+        help: "Version of the Tendrl managed sds, eg: '3.2.1'"
+        type: String
+    enabled: true
+
+  Cpu:
```
shtripat: Add help for all the attributes of other objects

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
       selinux_mode:
         type: String
 
+  Package:
```
shtripat: type missing for few attrs

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+        help: "Service state can be started|stopped|restarted|reloaded"
+        type: String
+      config_path:
+        help: "configuration file path for the service eg:/etc/tendrl/tendrl.conf"
+        type: String
+      config_data:
+        help: "Configuration data for the service"
+        type: String
+    atoms:
+      configure:
```
nthomas-redhat: does configure takes care of start and stop? we might have one for status as well

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+        uuid: "b90a0d97-8c9f-4ab1-8f64-dbb5638159a3"
+
+  Process:
+    enabled: True
+    attrs:
+      name:
+        help: "Name of the service"
+        type: String
+      state:
+        help: "Service state can be started|stopped|restarted|reloaded"
```
nthomas-redhat: Need to be updated

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+          - tendrl.node_agent.atoms.service.validations.check_service_running
+        run: tendrl.node_agent.atoms.service.configure.Configure
+        type: Update
+        uuid: "b90a0d97-8c9f-4ab1-8f64-dbb5638159a3"
+
+  Process:
+    enabled: True
+    attrs:
+      name:
+        help: "Name of the service"
```
nthomas-redhat: Need to be updated

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+        type: String
+      state:
+        help: "Service state can be started|stopped|restarted|reloaded"
+        type: String
+    atoms:
+      start:
```
nthomas-redhat: stop/status?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+        type: String
+      version:
+        help: "Version of the rpm/deb/pypi package"
+        type: String
+      pkg_type:
+        help: "Type of package can be rpm/deb/pip/"
+      state:
+        help: "State can installed|uninstalled"
+    atoms:
+      install:
```
nthomas-redhat: might require uninstall as well?

etc/tendrl_definitions_node_agent.yaml
```diff
@@ -83,3 +92,107 @@ object_details:
...
+        name: "Configure Service"
+        post_run:
+          - tendrl.node_agent.atoms.service.validations.check_service_running
+        run: tendrl.node_agent.atoms.service.configure.Configure
+        type: Update
+        uuid: "b90a0d97-8c9f-4ab1-8f64-dbb5638159a3"
+
+
+flows:
+  ImportGlusterCluster:
```
brainfunked: System specific flows should not be hardcoded in the main node agent flows.

Tendrl has to support multiple storage systems, but none of the support functionality should be hardcoded in any the tendrl components themselves. Specifically in import and create flows, no cluster context exists, so some tendrl support needs to be shipped with the node agents themselves, but this support functionality should be added as modules. This means that the node agent should be able to load the support modules at runtime, on a deployed system, via a predefined mechanism (convention over configuration).

The system specific support module would supply the following:
* system specific definition yaml for the various provisioning flows
* code that implements the specific atoms from these flows.

As an example of the definition file for this specific flow, something like [this gist](https://gist.github.com/brainfunked/1d8fd51ae5a4aa13d5b348530ed49c56)

Some notes regarding this file:
* One of the classes invoked from this definition file is `tendrl.node_agent.gluster_support.atoms.GenerateBridgeConfigurationFile`. This would be part of the code supplied by the gluster provisioning support module for the node agent.
* The gist also introduces the concept of namespaces to allow multiple yaml files to be grouped logically together at the parsing and compilation phase in the appropriate codebase (node agent, api etc.).
* This particular yaml file appends to the namespace under the node_agent namespace that the node_agent would expose. Validation checks should prevent clobbering of global namespaces such as the node_agent.
* The file introduces the concept of inline atom execution using the shell syntax _$()_ . This syntax can be modified if needed, but this functionality would be required in the future.

PR #31, opened 2016-11-11
Update ansible_module_runner.py

shtripat:
Fixed the name of the node_agent section referred from configuration.


shtripat:
@nnDarshan @r0h4n @nthomas-redhat please review

PR #48, opened 2016-11-15
Fix "node[]" to "Node[]"

r0h4n:


r0h4n:
It passes the tendrl schema 0.3 validations, "Node" is valid under tendrl.node_agent.objects

PR #50, opened 2016-11-15
Update import_cluster.py to send type for job

shtripat:


shtripat:
@r0h4n @nthomas-redhat please review

PR #51, opened 2016-11-15
Read cluster_id from Tendrl_context.

nnDarshan:
While posting sub-jobs for importing cluster. the cluster_id was picked properly. This patch fixes that


nnDarshan:
@r0h4n @nthomas-redhat @shtripat This fix is needed for un blocking multi node cluster import Please review and merge it soon

PR #60, opened 2016-11-17
Fix log_config_path variable under common.

nnDarshan:
Fixed log_config_path variable in tendrl.conf which had wrong key and value


nnDarshan:
@nthomas-redhat @shtripat @TimothyAsir Please review.

PR #61, opened 2016-11-18
Node-agent: Add option in tendrl-config file to specify package type

nnDarshan:
This patch adds a config option in tendrl.conf to specify the deployment
type. The possible options are "production" and "dev". For now this
deployment type is used to figure out whether to use rpm based
installation or pip based installation of packages during import cluster.

### DIFF
--- a/etc/tendrl/tendrl.conf.sample
+++ b/etc/tendrl/tendrl.conf.sample
@@ -16,3 +16,6 @@ log_level = DEBUG
 log_level = DEBUG
 tendrl_exe_file_prefix = /tmp/.tendrl_runner
 log_cfg_path = /etc/tendrl/node_agent_logging.yaml
+
+# deployment type can be either "production" or "dev"
+deployment_type = production
--- a/tendrl/node_agent/ceph_integration/flows/import_cluster.py
+++ b/tendrl/node_agent/ceph_integration/flows/import_cluster.py
@@ -2,7 +2,10 @@ import json
 import socket
 import uuid
 
+from tendrl.node_agent.config import TendrlConfig
 from tendrl.node_agent.flows.flow import Flow
+
+config = TendrlConfig()
 
 
 class ImportCluster(Flow):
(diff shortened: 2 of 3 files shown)


nnDarshan:
@nthomas-redhat @TimothyAsir @shtripat Could you please review the patch. This is needed for RPM based installation to Tendrl packages

tendrl/node_agent/ceph_integration/flows/import_cluster.py
```diff
@@ -27,8 +30,14 @@ def run(self):
                                            json.dumps(job))
         if self.node_id in node_list:
             self.parameters['fqdn'] = socket.getfqdn()
-            ceph = "git+https://github.com/Tendrl/ceph_integration.git@v1.0"
-            self.parameters['Package.name'] = ceph
+            if config.get("node_agent", "deployment_type") == "production":
```
brainfunked: All of this hardcoding is supposed to be in the definition YAML files, and not in the code.

tendrl/node_agent/gluster_integration/flows/import_cluster.py
```diff
@@ -24,9 +27,13 @@ def run(self):
                                            json.dumps(job))
         if self.node_id in node_list:
             self.parameters['fqdn'] = socket.getfqdn()
-            gluster = "git+https://github.com/Tendrl/gluster_integration.git" \
-                      "@v1.0"
-            self.parameters['Package.name'] = gluster
+            if config.get("node_agent", "deployment_type") == "production":
```
brainfunked: Same as in the ceph_integration case -- turn this into YAML and supply it as a common library method that allows package installation or git checkouts for provisioning components. The YAML flow would then invoke the library method with the appropriate parameters. These parameters (package names, repo paths, commands to be executed etc.) should be hardcoded in individual flows' YAML files.

nnDarshan:
Have raised an issue to track this https://github.com/Tendrl/node_agent/issues/65
will fix it as part of the refactoring

etc/tendrl/tendrl.conf.sample
```diff
@@ -16,3 +16,6 @@ log_level = DEBUG
 log_level = DEBUG
 tendrl_exe_file_prefix = /tmp/.tendrl_runner
 log_cfg_path = /etc/tendrl/node_agent_logging.yaml
+
+# deployment type can be either "production" or "dev"
+deployment_type = production
```
r0h4n: This is wrong, a deployer should never have to define the type of environment to select the type of packaging that tendrl will enable

tendrl/node_agent/ceph_integration/flows/import_cluster.py
```diff
@@ -27,8 +30,14 @@ def run(self):
...
+                self.parameters['Package.name'] = "tendrl-ceph-integration"
+                self.parameters['Package.pkg_type'] = "yum"
```
r0h4n: we need to detect rpm or deb support based on node_agent "node" or "system" objects

PR #64, opened 2016-11-21
Added disk details in hardware inventory

GowthamShanmugam:


tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+                  "Vendor":     "string",
+                  "DiskId":     "uuid"}, ...], ...}
+    '''
+    devlist = {}
+    dev_info = {}
+    rv = []
+    columes = 'NAME,KNAME,FSTYPE,MOUNTPOINT,UUID,PARTUUID,MODEL,SIZE,TYPE,' \
+              'PKNAME,VENDOR'
+    keys = columes.split(',')
+    lsblk = ("lsblk --all --bytes --noheadings --output='%s' --path --raw" %
```
brainfunked: This should mean an explicit dependency on the util-linux package in the rpm spec file. If this isn't already the case, please add this dependency to the spec, increase the release version and commit the spec file along with this pull request too.

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+                  "Type":       "type",
+                  "Used":       boolean,
+                  "SSD":        boolean,
+                  "Vendor":     "string",
+                  "DiskId":     "uuid"}, ...], ...}
+    '''
+    devlist = {}
+    dev_info = {}
+    rv = []
+    columes = 'NAME,KNAME,FSTYPE,MOUNTPOINT,UUID,PARTUUID,MODEL,SIZE,TYPE,' \
```
brainfunked: Should be `columns`.

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+              'PKNAME,VENDOR'
+    keys = columes.split(',')
+    lsblk = ("lsblk --all --bytes --noheadings --output='%s' --path --raw" %
+             columes)
+    cmd = Command({"_raw_params": lsblk})
+    out, err = cmd.start()
+    if out:
```
brainfunked: What happens if the command fails? Shouldn't there be an exception defined and raised appropriately?

GowthamShanmugam: i think forgot to log the error message

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
     return tendrl_context
 
 
+def getNodeDisk():
```
shtripat: Use snake_case for function names in python :)

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+    '''returns structure
+
+    {"nodename": [{"DevName":   "devicename",
```
shtripat: Lets keep field names in the dict as small and snake_case

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+    keys = columes.split(',')
+    lsblk = ("lsblk --all --bytes --noheadings --output='%s' --path --raw" %
+             columes)
+    cmd = Command({"_raw_params": lsblk})
+    out, err = cmd.start()
+    if out:
+        devlist = map(lambda line: dict(zip(keys, line.split(' '))),
+                      out['stdout'].splitlines())
+        parents = set([d['PKNAME'] for d in devlist if 'PKNAME' in d])
+        for d in devlist:
```
shtripat: There had been issues around different OS platforms with lsblk output. Specifically I remember on CentOS. Verify the scenario against RHEL, CentOS and fedora.

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +122,98 @@ def getTendrlContext():
...
+                in_use = False
+
+            d.update({'INUSE': in_use})
+            dev_info.update({d['KNAME']: d})
+
+    for disk in dev_info.values():
+        try:
+            u = list(bytearray(uuid.UUID(disk["UUID"]).get_bytes()))
+        except ValueError:
+            # TODO(log the error)
```
shtripat: log the error ;)

GowthamShanmugam: ok i will add log for this file :)

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: i will add the package util-linux  as dependency to solve this problem

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: snake case :) i dont know what it is, ok i will refer and change  it

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: darshan told me to remove this line because still we dont have any uuid format. So he told me to keep it as string.

GowthamShanmugam: And i also checked, for some uuid format this statement is not working

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: @brainfunked  i have checked in centos, lsblk is available by default.  i think it is not required

tendrl/node_agent/manager/pull_hardware_inventory.py
brainfunked: @GowthamShanmugam Even if the package exists by default, still add a dependency on it because you actually have a direct dependency on it. Always ensure that all direct dependencies are explicitly specified.

tendrl/node_agent/manager/pull_hardware_inventory.py
brainfunked: snake_case and CamelCase :)

tendrl/node_agent/manager/pull_hardware_inventory.py
brainfunked: @shtripat :+1: 

I would also add Ubuntu LTS to the list btw.

GowthamShanmugam: ok :)

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: i will add it

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: i will add it

GowthamShanmugam:
@brainfunked , @shtripat , @nnDarshan all changes are done

GowthamShanmugam:
@shtripat , @nnDarshan , @brainfunked , @r0h4n  pls review it

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +126,95 @@ def getTendrlContext():
...
+                    in_use = False
+            elif not d['FSTYPE']:
+                in_use = False
+
+            d.update({'INUSE': in_use})
+            dev_info.update({d['KNAME']: d})
+
+        for disk in dev_info.values():
+            if disk['TYPE'] == 'disk':
+                ssdStat = Is_ssd(disk['KNAME'])
```
shtripat: Rename thefunction as is_ssd

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +126,95 @@ def getTendrlContext():
...
+                       "fs_type": disk["FSTYPE"],
+                       "fs_uuid": disk["UUID"],
+                       "model": disk["MODEL"],
+                       "mount_point": [disk["MOUNTPOINT"]],
+                       "name": disk["NAME"],
+                       "parent": disk["PKNAME"],
+                       "size": long(disk["SIZE"]),
+                       "type": disk["TYPE"],
+                       "used": disk["INUSE"],
+                       "ssd": ssdStat,
```
shtripat: ssd_stat ??

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -137,5 +230,6 @@ def get_node_inventory():
     node_inventory["cpu"] = getNodeCpu()
     node_inventory["memory"] = getNodeMemory()
     node_inventory["tendrl_context"] = getTendrlContext()
+    node_inventory["disks"] = Get_node_disk()
```
shtripat: Rename the function as get_node_disks

tendrl/node_agent/persistence/disk.py
```diff
@@ -0,0 +1,31 @@
+from tendrl.common.etcdobj.etcdobj import EtcdObj
+from tendrl.common.etcdobj import fields
+
+
+class Disk(EtcdObj):
+    """A table of the Disk, lazily updated
+
+    """
+    __name__ = 'nodes/%s/Disk/%s'
```
shtripat: URl should be like nodes/<nodeid>/disks/<diskid> or nodes/<nodeid>/devices/<deviceid>

tendrl/node_agent/persistence/disk.py
```diff
@@ -0,0 +1,31 @@
...
+    type = fields.StrField("type")
+    used = fields.StrField("used")
+    ssd = fields.StrField("ssd")
+    vendor = fields.StrField("vendor")
+    disk_id = fields.StrField("disk_id")
+
+    def render(self):
+        self.__name__ = self.__name__ % (
+            self.node_id,
+            (self.device_name.replace("/", "", 1)).replace("/", "_")
```
shtripat: Shouldnt we disk_id as second arg in URL? Suggestions??

tendrl/node_agent/tests/test_disk.py
```diff
@@ -0,0 +1,92 @@
+from tendrl.node_agent.persistence.disk import Disk
+
+
+class Test_Disk(object):
```
shtripat: Avoid underscores in class names and rather use camelCase with first letter uppercase

tendrl/node_agent/tests/test_disk.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_Disk(self):
+        self.disk = Disk()
+        self.disk.device_name = "/temp/test"
```
shtripat: /dev/vdb etc are more natural names :))

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: ssd state .. if it is ssd then it contains True otherwise False

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: i will rename it

tendrl/node_agent/persistence/disk.py
GowthamShanmugam: nodes//disks/ why we have to use double slash (//)

tendrl/node_agent/persistence/disk.py
GowthamShanmugam: Disk id will come only when it is used, if the disk is not used then it does not have any id, so we can use disk name only

tendrl/node_agent/tests/test_disk.py
GowthamShanmugam: i will correct it

tendrl/node_agent/tests/test_disk.py
GowthamShanmugam: i will rename it

tendrl/node_agent/persistence/disk.py
GowthamShanmugam: as per your comment i will change url like (nodes/%s/Disk/%s)

GowthamShanmugam:
@shtripat , @nnDarshan , @brainfunked , @r0h4n  all changes are done

tendrl/node_agent/persistence/disk.py
```diff
@@ -0,0 +1,31 @@
...
+    node_id = fields.StrField("node_id")
+    device_name = fields.StrField("device_name")
+    fs_type = fields.StrField("fs_type")
+    fs_uuid = fields.StrField("fs_uuid")
+    model = fields.StrField("model")
+    mount_point = fields.StrField("mount_point")
+    name = fields.StrField("name")
+    parent = fields.StrField("parent")
+    size = fields.StrField("size")
+    type = fields.StrField("type")
```
nnDarshan: "type" is a python builtin, so please dont use it as variable, have something like "disk_type". If you have used it anywhere else please replace it

GowthamShanmugam:
@r0h4n @shtripat @nnDarshan please review it

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,36 @@
           type: String
       enabled: true
       value: nodes/$Node_context.node_id/Memory
+    Disks:
```
r0h4n: it is "Disk" and not "Disks"

GowthamShanmugam: i will change it

GowthamShanmugam:
@nthomas-redhat @r0h4n @shtripat @nnDarshan please review it

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,36 @@
           type: String
       enabled: true
       value: nodes/$Node_context.node_id/Memory
+    Disk:
```
shtripat: Add `help` field for all the attributes..

tendrl/node_agent/manager/manager.py
```diff
@@ -227,6 +228,34 @@ def on_pull(self, raw_data):
                     node_id=raw_data["node_id"],
                 )
             )
+        if "disks" in raw_data:
+            try:
+                self.etcd_client.delete(
+                    ("nodes/%s/Disks") % raw_data["node_id"], recursive=True)
```
r0h4n: why delete directory "nodes//Disks" ?

tendrl/node_agent/manager/manager.py
```diff
@@ -227,6 +228,34 @@ def on_pull(self, raw_data):
...
+            except Exception as ex:
```
r0h4n: Dont catch generic exception and simply log it. It can be any sort of exception, find python-etcd specific exception and catch that https://github.com/jplana/python-etcd/blob/master/src/etcd/__init__.py#L118...L307

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +126,95 @@ def getTendrlContext():
...
+    cmd = Command({"_raw_params": temp})
+    out, err = cmd.start()
+    if not out:
+        LOG.error(err)
+        raise Exception("Failed to get cluster statistics")
+    if out == '0':
+        return True
+    if out == '1':
+        return False
+    """Rotational attribute not found for
```
r0h4n: Wrongly Copied from : https://github.com/skyrings/skyring/blob/master/backend/salt/python/skyring/saltwrapper.py#L543

Is this supposed to be a code comment?

tendrl/node_agent/persistence/disk.py
```diff
@@ -0,0 +1,39 @@
...
+        if self.used == "True":
+            self.__name__ = self.__name__ % (
+                self.node_id,
+                "used",
+                self.disk_id
+            )
+        else:
+            self.__name__ = self.__name__ % (
+                self.node_id,
+                "free",
```
r0h4n: what is the meaning of free here?

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,36 @@
...
+        used:
+          type: Boolean
+        ssd:
+          type: Boolean
+        vendor:
+          type: String
+        disk_id:
+          type: String
+      enabled: true
+      value: nodes/$Node_context.node_id/Disks/device_name
```
r0h4n: In file https://github.com/Tendrl/node_agent/pull/64/files#diff-ce609b2cc94f6f7d7cd1d48f43d75492 you have defined disk etcd layout as "nodes/%s/Disks/%s/%s" 

and here it is "nodes/$Node_context.node_id/Disks/device_name", which one is correct?

tendrl/node_agent/manager/manager.py
GowthamShanmugam: Lets take example, node have two disk vdb and vdc. When node_agent start these two disk details are stored in etcd. if vdb removed, then vdc details only updated and vdb details are not affected in etcd. Because as per etcd key details goes to vdc only. No vdb details comes, so vdb never affected. To solve this problem when ever any change comes in local inventory file and node inventory list, I am just deleting disk details and storing fresh details in etcd. So only vdc disk details are in etcd.

tendrl/node_agent/manager/manager.py
GowthamShanmugam: When node agent start first time, there is no disk details in etcd. Code for push disk details is under this code. If it try to delete disk details from etcd, there no disk details pushed in etcd yet, so it raise exception and Because of that exception insert code never execute and control came out from function. This is the reason i catch exception here. Next time etcd details are present in etcd so it will delete old details and new disk details are pushed.

tendrl/node_agent/persistence/disk.py
GowthamShanmugam: I am storing free nodes separately and used node separately
free nodes comes under : nodes/node_id/Disks/free/disk_name
 used nodes comes under : nodes/node_id/Disks/used/disk_id

tendrl/node_agent/manager/manager.py
r0h4n: Makes sense

tendrl/node_agent/manager/manager.py
r0h4n: You are catching the broadest python exception "Exception" and simply logging it and it gets suppressed , this is not allowed usually, read more here 

https://realpython.com/blog/python/the-most-diabolical-python-antipattern/

r0h4n: Figure our the broadest exception thrown by python-etcd when a given key is not found in etcd and catch it

tendrl/node_agent/manager/pull_hardware_inventory.py
r0h4n: Is it not a important comment, why remove it?

tendrl/node_agent/persistence/disk.py
r0h4n: Are you aware of how "value" attribute is used by API to GET object details?
https://github.com/Tendrl/specifications/blob/master/specs/use_value_attribute_to_locate_object_paths_in_etcd.adoc

You are breaking that functionality, please read the spec and fix your PR

GowthamShanmugam: Api for this i am going give like nodes/$Node_context.node_id/Disks

tendrl/node_agent/manager/manager.py
GowthamShanmugam: ok

tendrl/node_agent/manager/pull_hardware_inventory.py
GowthamShanmugam: sorry i am not read that comment clearly...... What that comment says means,  The command will give value like 0 or 1, If the disk is ssd it will give 0 , if it is hard disk then it give 1, If any other value came or nothing is came then we can say this is ssd or hard disk.  

This comment is required, it gives correct information.

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
GowthamShanmugam: i update both place like nodes/$Node_context.node_id/Disks

tendrl/node_agent/manager/pull_hardware_inventory.py
r0h4n: makes sense, resolved.

tendrl/node_agent/persistence/disk.py
r0h4n: It is a problem, how is the API supposed to understand that it has to check in "Free" and "Used" directories?

tendrl-api just uses the "list" attribute of an object to list its id's and the "value" attribute of an object to fetch specific IDs.

GowthamShanmugam: Can you please tell me how i have to give, It is confusing me

GowthamShanmugam: anup suggested this list: nodes/$Node_context.node.id/Disks is it work?

tendrl/node_agent/manager/manager.py
shtripat: Similar thing we used to do in skyring earlier as well to overcome the issue

tendrl/node_agent/persistence/disk.py
shtripat: I get the issue @r0h4n is talking about. But I understand API or UI layer needs list of un-unsed disks in most the cases like create cluster flow, add brick/osd flow etc.

So I feel we should have two different list APIs for used and free disks. But not sure how this would be achieved under one object definition only. I dont think we can have two `list` values defined for an object.

@r0h4n @brainfunked thoughts??

GowthamShanmugam:
@shtripat @r0h4n @nnDarshan @nthomas-redhat please review

tendrl/node_agent/manager/pull_hardware_inventory.py
```diff
@@ -122,6 +126,105 @@ def getTendrlContext():
...
+
+            d.update({'INUSE': in_use})
+            dev_info.update({d['KNAME']: d})
+            cmd = ("hdparm -I %s" % d['KNAME'])
+            cmd = Command({"_raw_params": cmd})
+            out, err = cmd.start()
+            if not out['stderr']:
+                out = out['stdout']
+                info_list = out.split('\n')
+                d['SERIAL'] = info_list[5].split(':')[1].strip()
```
brainfunked: This is extremely fragile. It relies on hdparm outputting a specific information in a very specific sequence. At least parse for a specific string and then extract the information. Even in that case, it is still fragile, but nowhere near as bad as this. And don't use an exact string match, just use a string that's specific enough to give you the line that contains the information you want.

GowthamShanmugam: @brainfunked i am changed

GowthamShanmugam: @brainfunked please review

brainfunked:
I would suggest `hwinfo` be investigated. It has the ability to dump the output to a database. When invoked again, with an existing database, it is also capable of showing the list of changes made to the disk configuration.

GowthamShanmugam:
@brainfunked hwifo is also not giving any id for free nodes

GowthamShanmugam:
@brainfunked @shtripat @nthomas-redhat @nnDarshan please review

GowthamShanmugam:
@nthomas-redhat @brainfunked @shtripat review please

tendrl/node_agent/manager/manager.py
```diff
@@ -227,6 +230,35 @@ def on_pull(self, raw_data):
                     node_id=raw_data["node_id"],
                 )
             )
+        if "disks" in raw_data:
+            LOG.info("on_pull, Updating disks")
+            try:
+                self.etcd_client.delete(
```
shtripat: What is raw_disks passed is None? Do I feel we should delete the entry from etcd post some initial checks only.

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,169 @@
           type: String
       enabled: true
       value: nodes/$Node_context.node_id/Memory
+    disk:
```
shtripat: Object name to start with capital letter.

tendrl/node_agent/persistence/free_disk.py
```diff
@@ -0,0 +1,8 @@
+from tendrl.node_agent.persistence.disk import Disk
```
shtripat: I feel FreeDisk and UsedDisk classes could be added to disk.py only..

tendrl/node_agent/manager/manager.py
GowthamShanmugam: i got your point , i will delete disk when something comes in raw_disks

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
GowthamShanmugam: i will change it

tendrl/node_agent/persistence/free_disk.py
GowthamShanmugam: i will change it

tendrl/node_agent/manager/manager.py
GowthamShanmugam: The prob is if all the disk are removed then raw disk will come empty, at that time all disk should be deleted in etcd, so i feel this implementation works  fine

shtripat: ok

GowthamShanmugam:
@shtripat @nthomas-redhat @brainfunked please review

GowthamShanmugam:
@shtripat @nthomas-redhat @brainfunked please review

doc/source/installation.rst
```diff
@@ -39,6 +39,8 @@ bundeled for syslog and journald logging as well. These could be used similarly
 For installing node agent following dependencies have to be manually installed:
 
     $ yum install libffi-devel gcc python-devel openssl-devel
+    $ yum install -y http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-1.el7.nux.noarch.rpm
```
shtripat: I am not sure if absolute URL references are correct here.

tendrl/node_agent/manager/manager.py
```diff
@@ -227,6 +228,29 @@ def on_pull(self, raw_data):
...
+                    ("nodes/%s/Disks") % raw_data["node_id"], recursive=True)
+            except etcd.EtcdKeyNotFound as ex:
+                LOG.debug("Given key is not present in etcd . %s", ex)
+            disks = raw_data['disks']
+            if "disks" in disks:
+                for disk in disks['disks']:
+                    disk['node_id'] = raw_data['node_id']
+                    disk_obj = Disk(disk)
+                    disk_json = disk_obj.to_json_string()
```
shtripat: I understand you specifically convert to json and then write the Disk details to etcd. This you are doing because etcd object write does write field by field and there are multiple calls involved to etcd on that case. This used to be a performance issue as there are huge no of attributes associated with Disk object.
Just an info to others behind using this way of writing data.

tendrl/node_agent/persistence/disk.py
```diff
@@ -0,0 +1,69 @@
+import json
+import time
+
+
+class Disk(object):
```
shtripat: I understand the logic behind this class not being extended from etcdobject and rather a json representation of the same would be written one shot to etcd :)

doc/source/installation.rst
GowthamShanmugam: @shtripat No we have to build that package, so i have raise an issue for that also, upto that we can use this link for check right?

shtripat: ok

tendrl/node_agent/manager/manager.py
GowthamShanmugam: @shtripat yes this is the exact problem, that is the reason i am converting class object as json

shtripat:
LGTM. @nthomas-redhat @brainfunked please review.

tendrl/node_agent/persistence/disk.py
GowthamShanmugam: @shtripat yes we are using the same

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,172 @@
...
+        optimal_io_size:
+          help: "optimal I/O size"
+          type: String
+        log_sector_size:
+          help: "logical sector size"
+          type: String
+
+      enabled: true
+      list: nodes/$Node_context.node_id/Disks
+    UsedDisk:
```
r0h4n: Where have you implemented the "UsedDisk" and "FreeDisk" objects? I only see them defined here

r0h4n:
Please resolve merge conflicts

GowthamShanmugam:
@r0h4n conflicts resolved, please review

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
r0h4n: @GowthamShanmugam both these objects need to be defined in tendrl definitions, you cannot simply implement them and put them in etcd

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -30,6 +30,172 @@
...
+      attrs:
+        disk_id:
+          help: "File system UUID"
+          type: String
+      enabled: true
+      list: nodes/$Node_context.node_id/Disks/used
+    FreeDisk:
+      attrs:
+        disk_id:
+          help: "File system UUID"
```
GowthamShanmugam: @r0h4n object for free disk and used disk(above this)

tendrl/node_agent/manager/manager.py
```diff
@@ -214,6 +215,29 @@ def on_pull(self, raw_data):
...
+                    self.etcd_client.write((disk_obj.__name__) % (
+                        raw_data["node_id"], disk['disk_id']), disk_json)
+            if "used_disks_id" in disks:
+                for disk in disks['used_disks_id']:
+                    self.etcd_client.write(("nodes/%s/Disks/used/%s") % (
+                        raw_data["node_id"], disk), "")
```
GowthamShanmugam: @r0h4n code for used disk and free disk(below this)

r0h4n: @GowthamShanmugam I have repeatedly requested, implement classes for objects defined under Tendrl (class FreeDisk, class UsedDisk) and not simply write these objects to etcd, anyways Ill take this up in another issue, this is too time consuming

Issue #71, opened 2016-11-28
Fix the pre and post run atoms of import cluster flow


nnDarshan:
The pre and post run atoms of import cluster flow has to be fixed:
* The atom "compare" has to properly compare the sds name and version that is provided by the user as part of import cluster with the actual sds version and type that is installed in the nodes provided by user for import.
* The atom check-cluster-id-exists has to make sure that an entry of the cluster that was imported is available in the central store after completion of the import cluster flow(a post run atom).

PR #72, opened 2016-11-28
Node-agent: Fix the pre and post run atoms of import cluster flow.

nnDarshan:
This patch fixes the pre atom "compare" and the post atom
"chech_cluster_id_exists" of the import cluster flow.

Issue #71


nnDarshan:
@nthomas-redhat @shtripat @r0h4n

brainfunked:
At worst, this code needs to go into commons. Being able to check if a cluster id exists or not is an essential library method. This should not be part of the node agent codebase.

Secondly, this should be turned into a specification for framework enhancement before committing the code for it.

Value comparison, validations, checks for specific keys in etcd etc. needs to be part of the commons library. Generic methods that provide such functionality should be in commons. The inputs themselves should be picked up from the flow global variables, and the library atoms should invoked with these inputs via the flow files. This means that the flow files also need to support inputs to be provided to the atoms.

@nnDarshan @shtripat please turn this into a specification this week. Make a pull request on the specifications repository and we can review and expand the specification there.

shtripat:
@brainfunked https://tendrl.atlassian.net/browse/TEN-157 has listed various common module refactor stuff.
Can you please create required github issues and assign to respective people?

brainfunked:
@shtripat Can't create issues till there's a specification. We'll link the specification to that jira issue. The github issues will be created based on the specification.

PR #85, opened 2016-12-07
Fixes stale local Node_context

r0h4n:
- Compare local Node_context with central store Node_context
- Delete and re-generate local Node_context if above doesnt match

tendrl-bug-id: Tendrl/node_agent#82


r0h4n:
@Tendrl/tendrl-core @shtripat @nnDarshan please review

PR #92, opened 2016-12-08
convert README file into rst format

mbukatov:
tendrl-bug-id: Tendrl/node_agent#90


r0h4n:
If PyPi/python requires .rst docs we will have them. But we should not delete .adoc docs, there's no doubt that adoc is much better. And asciidoctor is supported on github

Please change the PR and keep both .rst and .adoc files

mbukatov:
@r0h4n rebase done

PR #95, opened 2016-12-08
Logged EtcdKeyNotFound exception from rpc.py

GowthamShanmugam:
tendrl-bug-id:Tendrl/node_agent#94

Signed-off-by: root <gshanmug@redhat.com>


GowthamShanmugam:
@r0h4n please review

PR #117, opened 2016-12-30
Invoke the integration process as service.

nnDarshan:
This patch starts the gluster and ceph integration processes
as service instead of background process. Also it writes the
cluster_id to a file, from where the integration services
reads the cluster_id.

tendrl-bug-id: Tendrl/node_agent#116

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include LICENSE
 include AUTHORS
-include README.adoc
+include README.rst
 include ChangeLog
 include requirements.txt
 include tendrl-node-agent.service
--- a/tendrl-node-agent.spec
+++ b/tendrl-node-agent.spec
@@ -30,9 +30,6 @@ Python module for Tendrl node bridge to manage storage node in the sds cluster
 
 %prep
 %setup
-# Remove the requirements file to avoid adding into
-# distutils requiers_dist config
-rm -rf {test-,}requirements.txt
 
 # Remove bundled egg-info
 rm -rf %{name}.egg-info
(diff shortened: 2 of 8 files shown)


nnDarshan:
@shtripat @nthomas-redhat @r0h4n Please review

tendrl/node_agent/objects/service/atoms/start.py
```diff
@@ -0,0 +1,25 @@
...
+    import service
+
+TENDRL_CLUSTER_ID_PATH_PREFIX = "/var/lib/tendrl/"
+
+
+class Start(object):
+    def run(self, parameters):
+        # write the cluster_id to cluster_id file
+        cluster_id_path = TENDRL_CLUSTER_ID_PATH_PREFIX + parameters.get(
+            "Tendrl_context.sds_name") + "_cluster_id"
```
anmolbabu: My feeling is that, this atom would be more useful if it just performed the task of starting a service(any service). 

My understanding is that, atoms are not directly exposed to outside world, its only that they are exposed via a flow and that IDEALLY is exposed ONLY via an api.

So, given that, can this atom be made generic to start any service and only start any service(can the cluster id part move else where), as I see a good use of such a generic atom. Else, if such a generic module already exists then shouldn't this be the like the flow does in its 
* pre section, the cluster id storage into a file part
and probably start the service using the generic already existing module as part of pre or post sections as applicable? Just that I am not sure if this atom is actually atomic :)

Note: Correct me if my understanding of the exposure of atom is incorrect.

nnDarshan:
Config has to be pushed to the common util for service invocation

PR #123, opened 2017-01-05
Populate the current node's id as member in manager

shtripat:
tendrl-bug-id: Tendrl/node_agent#122
tendrl-spec: multiple_node_ids_in_jobs.adoc

### DIFF
--- a/tendrl/node_agent/manager/manager.py
+++ b/tendrl/node_agent/manager/manager.py
@@ -100,17 +100,18 @@ class NodeAgentManager(Manager):
                     is None:
                 utils.delete_local_node_context()
 
+        node_id = utils.get_local_node_context()
         super(
             NodeAgentManager,
             self
         ).__init__(
             "node",
-            utils.get_local_node_context(),
-            utils.get_local_node_context(),
+            node_id,
             config,
             NodeAgentSyncStateThread(self),
             NodeAgentEtcdPersister(config),
-            "/tendrl_definitions_node-agent/data"
+            "/tendrl_definitions_node_agent/data",
+            node_id=node_id,
         )
         self.register_node(machine_id)
 
--- a/tendrl/node_agent/tests/test_manager.py
+++ b/tendrl/node_agent/tests/test_manager.py
@@ -35,7 +35,7 @@ class TestNodeAgentManager(object):
             self.manager._discovery_thread,
             manager.SyncStateThread
         )
-        assert self.manager.defs_dir == "/tendrl_definitions_node-agent/data"
+        assert self.manager.defs_dir == "/tendrl_definitions_node_agent/data"
 
     def test_register_node(self):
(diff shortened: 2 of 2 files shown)


shtripat:
@Tendrl/tendrl-core please review

tendrl/node_agent/manager/manager.py
```diff
@@ -105,6 +105,7 @@ def __init__(self, machine_id):
         ).__init__(
             "node",
             utils.get_local_node_context(),
+            utils.get_local_node_context(),
```
GowthamShanmugam: @shtripat Why it you are calling  utils.get_local_node_context() twise ?

shtripat: One is used as integration_context and other as node_id. They are used for two different purposes and as Manager constructor needs both, using it twice to pass same value for two different purposes.

nnDarshan:
Change looks good, please fix the unit test failures

tendrl/node_agent/manager/manager.py
GowthamShanmugam: @shtripat i feel instead of doing multiple file reading, read once and store that in var and reuse the var twice. is that ok?

tendrl/node_agent/manager/manager.py
```diff
@@ -99,12 +99,14 @@ def __init__(self, machine_id):
...
 
+        node_id = utils.get_local_node_context()
         super(
             NodeAgentManager,
             self
         ).__init__(
             "node",
-            utils.get_local_node_context(),
+            node_id,
+            node_id,
```
r0h4n: passing same node_id twice?

shtripat: Here once its passed as integration_id value and once as node_id. @GowthamShanmugam earlier had the same question.

r0h4n: Well its redundant, and integration_id is meant to be generated by the tendrl-api and it is definitely not node_id

shtripat: I understand the logic for integartion_id but in case of node_agent for the time being we are using node_id as integration_id and in case of gluster/ceph inetgartion cluster_id is used as integration_id. Also as the base Manager class needs both `integration_id` and `node_id` as constructor argument, so at the moment passing node_id for both.

r0h4n: You can set default values for arguments in python, that way you dont have to just pass around values for the sake of it. Please use that approach

shtripat: I was thinking of that. Only in case of node agent actually we need node_id and gluster/ceph integration actually dont need this. Will change the code accordingly and update the PRs.

shtripat:
@r0h4n we need to merge https://github.com/Tendrl/specifications/pull/104 for this PR to be merged. Travis is related to that.
I have resolved the test case failures.

PR #125, opened 2017-01-05
streamline new version release process

ktdreyer:
* Standardize the existing version numbers at "1.1", since that is the latest version in Git.

* Add a `__version__` number to node_agent's `__init__.py`, so that we can determine the module's version number at runtime.

* Reduce the multiple locations where the version number was hard-coded, so that we only have two: one in the rpm .spec, and one in `__version__`. (The rpm .spec hard-coded version number is to permit direct mock-scm builds with Copr.)

* Add code in `setup.py` to dynamically set the version number with new `bumpversion` and `release` sub-commands.

* Document the new release process and commands in `README.rst`


mbukatov:
Thanks for this proposal.

I have additional question: do you [specify dev versions of your projects like `1.2.0.dev1`](https://packaging.python.org/distributing/#standards-compliance-for-interoperability) with this setup? Right now, we don't do that, but I was wondering whether we should start doing that so that the latest dev build would not end up with the same version as the previous one or the last stable release. I mention this here since your change would require coordination across all python Tendrl project, so it's a perfect time to draft some versioning guidelines at the same time.

ktdreyer:
@mbukatov I've pushed an additional change to this branch today that provides a new `make snapshot` target.

This command will rewrite the package's version numbers according to Git. For example, the resulting SRPM will be `tendrl-node-agent-1.1.dev42-42.919fa51.src.rpm`, and the `__version__` in Python will be `1.1dev42`.

The idea is that a CI system (Jenkins) can run `make snapshot` after every Git commit to master, and then build the resulting .src.rpm, like so:

```
  copr-cli build tendrl/tendrl *.src.rpm
```

setup.py
```diff
@@ -11,9 +19,97 @@ def extract_requirements(filename):
...
+
+
+class BumpCommand(Command):
+    """ Bump the __version__ number and commit all changes. """
+
+    user_options = [('version=', 'v', 'version number to use')]
+
+    def initialize_options(self):
+        new_version = metadata['version'].split('.')
+        new_version[-1] = str(int(new_version[-1]) + 1)  # Bump the final part
```
r0h4n: Seems like this line updates the minor version only?

ktdreyer: Right, it bumps the minor version by default. You can use `--version` if you want to do a larger version bump.

setup.py
```diff
@@ -27,5 +123,6 @@ def extract_requirements(filename):
         'console_scripts': [
             'tendrl-node-agent = tendrl.node_agent.manager.manager:main'
         ]
-    }
+    },
+    cmdclass={'bump': BumpCommand, 'release': ReleaseCommand},
```
r0h4n: Id prefer "bumpversion" instead of "bump"

ktdreyer: I think `bump` is faster to type, but I've changed it to `bumpversion` now so we can continue :)

PR #129, opened 2017-01-10
Fix multi-node cluster import failure

nnDarshan:
This patch fixes multi-node import failure
that is currently happening because we are
trying to dump a json that has a python
object.

tendrl-bug-id: Tendrl/node_agent#128


nnDarshan:
@r0h4n @nthomas-redhat @shtripat

PR #134, opened 2017-01-12
Added SDS detection modules

shtripat:
tendrl-bug-id: Tendrl/specifications#113
tendrl-spec: sds_detection_modules.adoc

Signed-off-by: Shubhendu <shtripat@redhat.com>

PR #135, opened 2017-01-13
Platform detection modules

nthomas-redhat:
tendrl-bug-id: Tendrl/specifications#112
tendrl-spec: Node agent system detection and integration modules

### DIFF
--- a/tendrl/node_agent/manager/manager.py
+++ b/tendrl/node_agent/manager/manager.py
@@ -15,8 +15,6 @@ from tendrl.commons.manager.manager import Manager
 from tendrl.commons.manager.manager import SyncStateThread
 from tendrl.node_agent.persistence.tendrl_definitions import TendrlDefinitions
 
-config = TendrlConfig("node-agent", "/etc/tendrl/tendrl.conf")
-
 from tendrl.node_agent.manager.tendrl_definitions_node_agent import data as \
     def_data
 from tendrl.node_agent.manager import utils
@@ -26,9 +24,13 @@ from tendrl.node_agent.persistence.memory import Memory
 from tendrl.node_agent.persistence.node import Node
 from tendrl.node_agent.persistence.node_context import NodeContext
 from tendrl.node_agent.persistence.os import Os
+from tendrl.node_agent.persistence.platform import Platform
 from tendrl.node_agent.persistence.persister import NodeAgentEtcdPersister
 from tendrl.node_agent.persistence.tendrl_context import TendrlContext
+from tendrl.node_agent.discovery.platform.manager import PlatformManager
+from tendrl.node_agent.discovery.platform.base import PlatformDiscoverPlugin
 
+config = TendrlConfig("node-agent", "/etc/tendrl/tendrl.conf")
 LOG = logging.getLogger(__name__)
 HARDWARE_INVENTORY_FILE = "/etc/tendrl/tendrl-node-inventory.json"
 
@@ -100,19 +102,21 @@ class NodeAgentManager(Manager):
                     is None:
                 utils.delete_local_node_context()
 
+        node_id = utils.get_local_node_context()
         super(
             NodeAgentManager,
             self
         ).__init__(
             "node",
-            utils.get_local_node_context(),
-            utils.get_local_node_context(),
+            node_id,
             config,
             NodeAgentSyncStateThread(self),
             NodeAgentEtcdPersister(config),
-            "/tendrl_definitions_node-agent/data"
-        )
+            "/tendrl_definitions_node_agent/data",
+            node_id=node_id,
+            )
         self.register_node(machine_id)
+        self.load_and_execute_platform_discovery_plugins()
 
     def register_node(self, machine_id):
         self.persister_thread.update_node_context(
@@ -238,6 +242,36 @@ class NodeAgentManager(Manager):
                 for disk in disks['free_disks_id']:
                     self.etcd_client.write(("nodes/%s/Disks/free/%s") % (
                         raw_data["node_id"], disk), "")
+
+    def load_and_execute_platform_discovery_plugins(self):
+        # platform plugins
+        LOG.info("load_and_execute_platform_discovery_plugins, platform \
+         plugins")
+        try:
+            pMgr = PlatformManager()
+        except ValueError as ex:
+            LOG.error(
+                'Failed to init PlatformManager. \Error %s' % str(ex))
+            return
+        # execute the platform plugins
+        for plugin in pMgr.get_available_plugins():
+            platform_details = plugin.discover_platform()
+            if len(platform_details.keys()) > 0:
+                # update etcd
+                try:
+                    self.persister_thread.update_platform(
+                        Platform(
+                            updated=str(time.time()),
+                            os=platform_details["Name"],
+                            os_version=platform_details["OSVersion"],
+                            kernel_version=platform_details["KernelVersion"],
+                            node_id=utils.get_local_node_context(),
+                        )
+                    )
+                except etcd.EtcdException as ex:
+                    LOG.error(
+                        'Failed to update etcd . \Error %s' % str(ex))
+                break
 
 
 def main():
--- a/tendrl/node_agent/manager/tendrl_definitions_node_agent.py
+++ b/tendrl/node_agent/manager/tendrl_definitions_node_agent.py
@@ -94,7 +94,7 @@ namespace.tendrl.node_agent:
           type: String
         bios_id:
           help: "Bios id"
-          type: String 
+          type: String
         state:
           help: "disk state"
           type: String
@@ -382,6 +382,16 @@ namespace.tendrl.node_agent:
           help: "configuration file path"
           type: String
       enabled: true
+    Platform:
+      attrs:
+        kernel_version:
+          type: String
+        os:
+          type: String
+        os_version:
+          type: String
+      enabled: true
+      value: nodes/$Node_context.node_id/Platform
 namespace.tendrl.node_agent.gluster_integration:
   flows:
     ImportCluster:
--- a/tendrl/node_agent/discovery/platform/base.py
+++ b/tendrl/node_agent/discovery/platform/base.py
@@ -1,0 +1,11 @@
+from abc import abstractmethod
+import six
+from tendrl.node_agent.discovery.plugin import PluginMount
+
+
+@six.add_metaclass(PluginMount)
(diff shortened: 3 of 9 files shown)


tendrl/node_agent/discovery_modules/platform/base.py
```diff
@@ -0,0 +1,15 @@
+from abc import abstractmethod
+import six
+from tendrl.node_agent.discovery_modules.plugin import PluginMount
+
+
+@six.add_metaclass(PluginMount)
+class PlatformBasePlugin(object):
+
+    @abstractmethod
+    def intialize(self):
```
shtripat: Do we need this function as we are not using for any means?

tendrl/node_agent/discovery_modules/platform/manager.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import importlib
+import inspect
+from tendrl.node_agent.discovery_modules.platform import base
+
+
+LOG = logging.getLogger(__name__)
+
+
+class PlatformManager(object):
```
shtripat: We wont be able to use common detection manager as we talked about. The plugins use meta class PluginMount and that maintains a list of plugins loaded as instance variable. If we maintain all the platform, sds etc plugins in same list, while picking and executing in node agent manager would be problematic.
We would end up writing different managers for different detection modules. Lets see, anyway I would write one for sds detection as well may be.

tendrl/node_agent/discovery_modules/platform/plugins/centos.py
```diff
@@ -0,0 +1,23 @@
+import logging
+import platform
+from tendrl.node_agent.discovery_modules.platform import base
+
+
+LOG = logging.getLogger(__name__)
+
+
+class CentosPlugin(base.PlatformBasePlugin):
```
shtripat: The name `PlatformBasePlugin` looks very generic. I feel we should call `PlatformDetectPlugin` or `PlatformDiscoverPlugin`

tendrl/node_agent/discovery_modules/platform/plugins/centos.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def intialize(self):
```
shtripat: Not required just for namesake.

tendrl/node_agent/discovery_modules/plugin.py
```diff
@@ -0,0 +1,11 @@
+class PluginMount(type):
```
shtripat: Dont re-define. I already have a commons util defined for this at https://github.com/Tendrl/commons/pull/102. Use the same.

tendrl/node_agent/manager/manager.py
```diff
@@ -100,19 +105,21 @@ def __init__(self, machine_id):
...
             config,
             NodeAgentSyncStateThread(self),
             NodeAgentEtcdPersister(config),
-            "/tendrl_definitions_node-agent/data"
-        )
+            "/tendrl_definitions_node_agent/data",
+            node_id=node_id,
+            )
         self.register_node(machine_id)
+        self.load_and_execute_discovery_plugins(node_id)
```
shtripat: Should execute this while loading the module or just before starting the manager in main()

tendrl/node_agent/manager/manager.py
```diff
@@ -239,6 +246,31 @@ def on_pull(self, raw_data):
                     self.etcd_client.write(("nodes/%s/Disks/free/%s") % (
                         raw_data["node_id"], disk), "")
 
+    def load_and_execute_discovery_plugins(self, node_id):
```
shtripat: I feel we can have generic module defined at tendrl.node_agent.discovery.manager.py which defines a function say `execute_discoveries()`. This class would define individual functions for loading different types of plugins and these functions would be called from inside the __init__() function of manager. Inside execute_discoveries we would execute each type of plugins one by one.

This class could look something like 
```
class DiscoveryManager(object):
    def __init__(self):
        try:
            self.load_platform_discovery_plugins()
            self.load_sds_discovery_plugins()
        except (SyntaxError, ValueError, ImportError) as ex:
            LOG.error('Failed to load the platform plugins. Error %s' %
                       ex, exc_info=True)
            raise ex
    
    def load_platform_discovery_plugins(self):
        ..........

    def load_sds_discovery_plugins(self):
        .............

    def execute_discoveries(self):
        for plugin in PlatformDiscoveryPlugin.plugins:
            # execute the function say discover_platform() and add details to etcd

        for plugin in SDSDiscoveryPlugin.plugins:
            # execute the function say discover_sds() and add details to etcd
```

From node agent manager all we need to do is 

`DiscoveryManager().execute_discoveries()` in `main()`

tendrl/node_agent/discovery_modules/platform/base.py
```diff
@@ -0,0 +1,15 @@
+from abc import abstractmethod
```
shtripat: I feel better module name is `tendrl.node_agent.discovery.platform`

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
```diff
@@ -382,6 +382,16 @@
           help: "configuration file path"
           type: String
       enabled: true
+    Platform:
+      attrs:
+        kernel_version:
+          type: String
+        os:
+          type: String
+        os_varsion:
```
nnDarshan: version ?

nnDarshan:
Can you please rebase the PR

tendrl/node_agent/manager/manager.py
shtripat: Check https://github.com/Tendrl/node-agent/pull/134. I have done the suggested changes here.

tendrl/node_agent/discovery_modules/plugin.py
nthomas-redhat: See the comment https://github.com/Tendrl/performance-monitoring/pull/23/commits/5a32480de1d7d369461601337683439803914255#r95658651
So it might pointless to keep it in 'commons'. So probably we can keep one for discovery plugins as long as it holds good

tendrl/node_agent/manager/manager.py
nthomas-redhat: I think this is the right place because the manager is initialized and the node is registered to the etcd

tendrl/node_agent/discovery_modules/platform/base.py
nthomas-redhat: I kept it as a place holder. I can remove it as well

tendrl/node_agent/discovery_modules/platform/plugins/centos.py
nthomas-redhat: ok

tendrl/node_agent/discovery_modules/platform/plugins/centos.py
nthomas-redhat: as mentioned above its a place holder

tendrl/node_agent/manager/tendrl_definitions_node_agent.py
nthomas-redhat: ok :)

tendrl/node_agent/discovery_modules/platform/base.py
nthomas-redhat: ok

tendrl/node_agent/discovery_modules/plugin.py
shtripat: I added in commons keeping in mind that in future if sds integration modules also introduce some plugins mechanism, this meta class could be used.

Also meta class doesnt restrict uses of only plugin but the way we get and execute the plugin. All that meta class provides is that we can load and maintain a list of plugins. Later stage using the interface name for the plugin we loop through the plugins and decide what to do.

Still I leave to team, may be we can have at discovery module level only.

tendrl/node_agent/manager/manager.py
shtripat: ok

tendrl/node_agent/manager/manager.py
```diff
@@ -239,6 +246,31 @@ def on_pull(self, raw_data):
...
+        # platform plugins
+        LOG.info("load_and_execute_discovery_plugins, platform plugins")
+        pMgr = PlatformManager()
+        pMgr.load_plugins()
```
GowthamShanmugam: During object creation of PlatformManager() a method load_plugins is called from __init__, why again this function is called?

tendrl/node_agent/manager/manager.py
```diff
@@ -4,6 +4,7 @@
 import signal
 import socket
 import time
+import etcd
```
GowthamShanmugam: Already etcd is imported

nthomas-redhat: done

tendrl/node_agent/manager/manager.py
nthomas-redhat: good catch, this is not required

tendrl/node_agent/manager/manager.py
nthomas-redhat: Having a single manger would be troublesome because all the plugins will be loaded into same array and you would have to run the unnecessary type checks while doing the discoveries, which adds up complexities. I would prefer to have this simple approach of having individual managers per modules

tendrl/node_agent/manager/manager.py
```diff
@@ -239,6 +243,35 @@ def on_pull(self, raw_data):
                     self.etcd_client.write(("nodes/%s/Disks/free/%s") % (
                         raw_data["node_id"], disk), "")
 
+    def execute_discovery_plugins(self, node_id):
+        # platform plugins
+        LOG.info("execute_discovery_plugins, platform plugins")
```
shtripat: log info to be like `LOG.info("load_and_execute_platform_discovery_plugins")`

tendrl/node_agent/manager/manager.py
```diff
@@ -239,6 +243,35 @@ def on_pull(self, raw_data):
                     self.etcd_client.write(("nodes/%s/Disks/free/%s") % (
                         raw_data["node_id"], disk), "")
 
+    def execute_discovery_plugins(self, node_id):
```
shtripat: Can be renamed as `load_and_execute_platform_discovery_plugins` and no need to pass `node_id` as parameter. We can always say `utils.get_local_node_context()`

PR #137, opened 2017-01-13
Add service detection feature to node agent

nnDarshan:
This patch adds service detection capability to node-agent.
Node-agent will have a list of predefined services, which
it will look out for on the node in which it is running. It
will update etcd with the list of services available and or
running on the node. Also based on if the service is running
it will update the TAGS field in the node_context.

tendrl-bug-id: Tendrl/node_agent#136

### DIFF
--- a/tendrl/node_agent/manager/manager.py
+++ b/tendrl/node_agent/manager/manager.py
@@ -8,6 +8,7 @@ import time
 import gevent.event
 import gevent.greenlet
 import pull_hardware_inventory
+from pull_service_status import TENDRL_SERVICE_TAGS
 
 from tendrl.commons.config import TendrlConfig
 from tendrl.commons.log import setup_logging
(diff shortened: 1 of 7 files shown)


nnDarshan:
@r0h4n @shtripat @nthomas-redhat Please review

nnDarshan:
Unit Tests not added as UT will fail until https://github.com/Tendrl/commons/pull/101 is merged

tendrl/node_agent/manager/manager.py
```diff
@@ -82,6 +84,18 @@ def _run(self):
         LOG.info("%s complete" % self.__class__.__name__)
 
 
+def update_node_context(manager, machine_id, tags=""):
```
nthomas-redhat: This should be in sync with https://github.com/Tendrl/node-agent/pull/134
one update should not overwrite the other

nnDarshan: This patch has been verified with, the patch you mentioned and it does not have any implication on that.

PR #143, opened 2017-01-16
Fix Configuration Loading related issues

nthomas-redhat:


etc/tendrl/node-agent/node-agent.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
shtripat: It should be a sample file and should be named as `node-agent.yaml.sample`

tendrl/node_agent/manager/manager.py
```diff
@@ -126,7 +125,7 @@ def __init__(self, machine_id):
             node_id,
             config,
             NodeAgentSyncStateThread(self),
-            NodeAgentEtcdPersister(config),
+            NodeAgentEtcdPersister(config, self.etcd_client),
```
shtripat: persister doesnt tale config as parameter anymore

nthomas-redhat:
@r0h4n @shtripat Please review

PR #144, opened 2017-01-17
Modified Node agent tags to be hierarchical.

nnDarshan:
This patch modifies node_agent tags to be hierarchical.


nnDarshan:
@shtripat @r0h4n @brainfunked @nthomas-redhat Please review. Node agent tags have been changed as per the discussion yesterday

tendrl/node_agent/manager/pull_service_status.py
```diff
@@ -1,14 +1,14 @@
...
 TENDRL_SERVICE_TAGS = {
-    "tendrl-node-agent": "TENDRL",
-    "etcd": "TENDRL_SERVER",
-    "tendrl-apid": "TENDRL_SERVER",
-    "tendrl-gluster-integration": "TENDRL_GLUSTER",
-    "tendrl-ceph-integration": "TENDRL_CEPH",
-    "glusterd": "GLUSTER",
-    "ceph-mon": "CEPH_MON",
-    "ceph-osd": "CEPH_OSD"
+    "tendrl-node-agent": "TENDRL/NODE",
```
shtripat: Should we really keep the names uppercase? I feel these would becomes keys in etcd as well. So we might make them lowercase for better typing.. suggestions?

nnDarshan: these are values in etcd. So this should fine I feel.

shtripat: ok

brainfunked: Let's downcase this. These would probably end up becoming keys when indexed.

nnDarshan: Okay, will change to lowercase

PR #147, opened 2017-01-17
[WIP] Tendrl object tree, Tendrl centralize definitions , Node Agent Flows and Namespaces

r0h4n:
- WIP: Tendrl object tree
- WIP: Tendrl centralize definitions
- WIP: Node Agent Flows and Namespaces

tendrl-bug-id: Tendrl/specifications#37
tendrl-bug-id: Tendrl/specifications#54
tendrl-bug-id: Tendrl/specifications#32

jira: [TEN-3] [TEN-4]


tendrl/node_agent/objects/base_object.py
```diff
@@ -0,0 +1,41 @@
...
+        self.attrs = attrs
+        self.enabled = enabled
+
+        # path to LIST of all instance of the object
+        self.obj_list = obj_list
+
+        # path to GET an instance of the object
+        self.obj_value = obj_value
+        self.atoms = atoms
+        self.flows = flows
```
shtripat: Where do we store global flows like CreateVolume and CreatePool which are not attached to an object?

shtripat:
@r0h4n are these changes verified for node_agent service? Incorporated the similar changes for ceph_integration component but not able to bring up the service. Shows the below error

```
Traceback (most recent call last):
  File "/usr/bin/tendrl-ceph-integration", line 11, in <module>
    load_entry_point('tendrl-ceph-integration==1.1', 'console_scripts', 'tendrl-ceph-integration')()
  File "/usr/lib/python2.7/site-packages/tendrl/ceph_integration/manager/manager.py", line 158, in main
    Config()
  File "/usr/lib/python2.7/site-packages/tendrl/ceph_integration/objects/config/Config.py", line 10, in __init__
    super(Config, self).__init__(*args, **kwargs)
TypeError: __init__() takes exactly 8 arguments (1 given)
```

Kindly help resolving the same.

tendrl/__init__.py
```diff
@@ -1 +1,33 @@
...
+        # Create the "Tendrl.node_agent.objects.$obj.{atoms, flows} NS
+    def add_object(self, obj_class, name):
+        # obj is the actual instance of that Tendrl object
+        # name of object as defined in Tendrl definitions
+        self.node_agent.objects[name] = obj_class
+
+    def add_flow(self, flow_class, name):
+        # flow is the actual instance of that Tendrl flow
+        # name of object as defined in Tendrl definitions
+        self.node_agent.flows[name] = flow_class
```
r0h4n: @shtripat here we store flows (global to tendrl.node_agent namespace), youll need to adapt BaseFlow as per BaseObject below

tendrl/node_agent/manager/manager.py
```diff
@@ -331,15 +297,23 @@ def load_and_execute_sds_discovery_plugins(self):
 
 
 def main():
+    # Register Config to tendrl.node_agent.objects namespace
+    Tendrl.config = Config()
```
shtripat: BaseObject takes no of arguments and we are not passing any here.

r0h4n: Check NodeAgentObject implementation here https://github.com/Tendrl/node-agent/pull/147/files#diff-5099f3ef3dbc5e1a16a38ce7dfbea8adR45

tendrl/node_agent/manager/manager.py
```diff
@@ -331,15 +297,24 @@ def load_and_execute_sds_discovery_plugins(self):
...
+        Tendrl.node_agent.objects.Config.data['log_level']
     )
-# init definitions
-# init manager and object tree from etcd
-    machine_id = utils.get_machine_id()
 
-    m = NodeAgentManager(machine_id)
+    etcd_kwargs = {'port': config['etcd_port'],
+                   'host': config["etcd_connection"]}
+    Tendrl.etcd_orm = etcdobj.Server(etcd_kwargs=etcd_kwargs)
```
anmolbabu: One experience that I have had while working on monitoring related stuff is when the same etcd client is used for multiple operations to/from etcd, specially in case of blocking watch operation, if a subsquent read/write operation is to be made, using the same etcd_client object, both fail, so, I feel it might not be good to make global instance of etcd_orm because that might fail in case of concurrent etcd accesses.

r0h4n: These are some steps to mitigate that.
- All writes to go via the persister thread as it was intended since the start.
- Reads can happen anywhere in the code but via the etcd_orm.read()
- One can tinker with the etcd_client's "per_host_pool_size" and "lock_prefix" to get more perf
https://github.com/jplana/python-etcd/blob/master/src/etcd/client.py#L66

- As for the "watch", it shouldn have been done via the main etcd_client anyways. Watch doesnt block the actual etcd api, if you simply use a temporary etcd_client instance for all your watches, it should be ok

tendrl/node_agent/objects/tendrl_context/TendrlContext.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.value = 'nodes/%s/TendrlContext'
+
+        # integration_id is the Tendrl generated cluster UUID
+        self.integration_id = integration_id or self._get_integration_id()
+        self.node_id = node_id
+
+    def save(self, persister):
+        cls_etcd = etcd_utils.to_etcdobj(_TendrlContextEtcd, self)
+        persister.save_node_context(cls_etcd())
```
nthomas-redhat: should this be save_tendrl_context?

r0h4n: why so?

r0h4n:
@nthomas-redhat @shtripat @nnDarshan @anmolbabu Please review and make corresponding changes to {ceph,gluster}-integration/node-agent/alerting/performance-monitoring

PR #151, opened 2017-01-19
create the detected_cluster object and populate etcd instead adding the details as part node_context

nthomas-redhat:
tendrl-bug-id: Tendrl/node-agent/issues#150
tendrl-spec : Node agent system detection and integration modules


r0h4n:
Looks good, gonna put this on hold and merge once all objects are in place. Thanks

PR #179, opened 2017-01-24
WIP: Fixing import gluster cluster

nnDarshan:


nnDarshan:
@r0h4n @nthomas-redhat @shtripat Please take a look. This is a WIP patch to fix import gluster cluster import flow. Currently with this PR, the control is going upto run() method of baseflow class, and is failing while executing atoms.

If anyone is free you can work on top of this.

nthomas-redhat:
Need to implement a generic import cluster flow and accept DetectedCluster as input in that and then based on detected cluster, call existing specific gluster or ceph import flows. In any case, older import flows should be working. can @nnDarshan  and @shtripat pair up and work on this?

shtripat:
@nnDarshan lets try to make existing import gluster and ceph cluster flows working first. In the mean time I would start on a generic flow for import cluster which would take detected cluster details as input and accordingly route to the required flows for actual import.

PR #190, opened 2017-01-27
Fix Unavailable gevent.signal.SIGTERM issue

nnDarshan:
Get SIGTERM value form signal module instead of
gevent.signal. Its not available in later


nnDarshan:
@r0h4n @brainfunked @nthomas-redhat @shtripat Please review

Issue #185, opened 2017-01-26
tendrl-node-agent service is not working


fbalak:
There is an error when I try run tendrl-node-agent after installation:

```
# systemctl status tendrl-node-agent
● tendrl-node-agent.service - A python agent local to every managed storage node in the sds cluster
   Loaded: loaded (/usr/lib/systemd/system/tendrl-node-agent.service; enabled; vendor preset: disabled)
   Active: failed (Result: start-limit) since Čt 2017-01-26 12:29:52 UTC; 1h 15min ago
 Main PID: 27912 (code=exited, status=1/FAILURE)

led 26 12:29:52 hostname systemd[1]: tendrl-node-agent.service: main process exited, code=exited, status=1/FAILURE
led 26 12:29:52 hostname systemd[1]: Unit tendrl-node-agent.service entered failed state.
led 26 12:29:52 hostname systemd[1]: tendrl-node-agent.service failed.
led 26 12:29:52 hostname systemd[1]: tendrl-node-agent.service holdoff time over, scheduling restart.
led 26 12:29:52 hostname systemd[1]: start request repeated too quickly for tendrl-node-agent.service
led 26 12:29:52 hostname systemd[1]: Failed to start A python agent local to every managed storage node in the sds cluster.
led 26 12:29:52 hostname systemd[1]: Unit tendrl-node-agent.service entered failed state.
led 26 12:29:52 hostname systemd[1]: tendrl-node-agent.service failed.
```

fbalak:
Tested with tendrl-node-agent-1.2-01_25_2017_21_31_48.noarch

fbalak:
Tested with:
tendrl-commons-1.2-01_30_2017_17_41_43.noarch
tendrl-node-agent-1.2-01_30_2017_17_13_12.noarch
tendrl-gluster-integration-1.2-01_30_2017_17_17_12.noarch

It looks ok.

PR #191, opened 2017-01-27
Fix tendrl-node-agent service is not working issue

TimothyAsirJeyasing:
tendrl-bug-id: Tendrl/node-agent#185

### DIFF
--- a/tendrl-node-agent.service
+++ b/tendrl-node-agent.service
@@ -3,7 +3,7 @@ Description= A python agent local to every managed storage node in the sds clust
 
 [Service]
 Type=simple
-Environment="HOME=default_foo_value"
+Environment="HOME=/var/lib/tendrl"
 ExecStart=/usr/bin/tendrl-node-agent
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=process
--- a/tendrl-node-agent.service
+++ b/tendrl-node-agent.service
@@ -3,6 +3,7 @@ Description= A python agent local to every managed storage node in the sds clust
 
 [Service]
 Type=simple
+Environment="HOME=default_foo_value"
 ExecStart=/usr/bin/tendrl-node-agent
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=process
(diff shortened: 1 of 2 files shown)


TimothyAsirJeyasing:
@brainfunked , @r0h4n , @nthomas-redhat , @nnDarshan , Please review

tendrl-node-agent.service
```diff
@@ -3,11 +3,12 @@ Description= A python agent local to every managed storage node in the sds clust
 
 [Service]
 Type=simple
+Environment="HOME=default_foo_value"
```
shtripat: Doesnt it look like some random value assigned? What is `default_foo_value` ??

brainfunked: What's with the value for $HOME here? Also, how was it figured out that $HOME is required?

TimothyAsir: The Home will here get the path /default_foo_value and the ansible_exec.... script file will be created under /default_foo_value/.tendrl/node-agent/ansible_exec..

I could not able to set the user home directory to the HOME variable. Because,
unfortunately systemd  requires full path and it could not able to expand $HOME.
for the environment variable. 

This issue can be fixed by either hard-code the value in our /etc//tendrl/node-agent/node-agent.conf.yaml file to use "/var/tmp" or we should set /root to HOME variable in service file if root is the only user going to execute.

Need your suggestion which way we can proceed.

nthomas-redhat: Set the $HOME to /var/lib/tendrl as discussed

TimothyAsir: done

TimothyAsir:
@brainfunked , @nthomas-redhat , @shtripat ,  @nnDarshan , Please review

tendrl-node-agent.service
```diff
@@ -3,11 +3,12 @@ Description= A python agent local to every managed storage node in the sds clust
 
 [Service]
 Type=simple
+Environment="HOME=/var/lib/tendrl"
```
brainfunked: Is the directory actually being created in the RPM spec file, however?

TimothyAsir: This directory created automatically during service start during ansible execution.
However for safer side, i will do changes in the spec file to create this directory.

PR #195, opened 2017-01-31
Multi node cluster import failure

nthomas-redhat:
tendrl-bug-id: Tendrl/node-agent/issues#194


sankarshanmukhopadhyay:
Over-riding the failures after a conversation with @nthomas-redhat and merging.

Issue #193, opened 2017-01-31
Detected cluster is not populated for osds


nthomas-redhat:
ceph -s -f json and ceph version -f json won't work on osds.
Rather you can use ceph --version command. For getting fsid, an option could be reading /etc/ceph/<cluster>.conf file

shtripat:
@nthomas-redhat plz close this as required PR is merged.

PR #196, opened 2017-01-31
Corrected sds detection logic for OSD nodes

shtripat:
tendrl-bug-id: Tendrl/node-agent#193

### DIFF
--- a/tendrl/node_agent/discovery/sds/plugins/discover_ceph_sds_details.py
+++ b/tendrl/node_agent/discovery/sds/plugins/discover_ceph_sds_details.py
@@ -32,10 +32,28 @@ class DiscoverCephStorageSystem(DiscoverSDSPlugin):
             ret_val['pkg_name'] = details[0]
 
             # get the cluster_id details
-            raw_data = ini2json.ini_to_dict("/etc/ceph/ceph.conf")
-            if "global" in raw_data:
-                ret_val['detected_cluster_id'] = raw_data['global']['fsid']
-                ret_val['cluster_attrs'] = {'fsid': raw_data['global']['fsid'],
-                                            'name': 'ceph'}
+            os_name = tendrl_ns.platform.os
+            cfg_file = ""
+            if os_name in ['CentOS Linux', 'Red Hat Enterprise Linux Server']:
+                cfg_file = '/etc/sysconfig/ceph'
+            #TODO(shtripat) handle the case of ubuntu
+
+            if cfg_file != "":
+                with open(cfg_file) as f:
+                    for line in f:
+                        if line.startswith("CLUSTER="):
+                            cluster_name = line.split('\n')[0].split('=')[1]
+
+            if cluster_name:
+                raw_data = ini2json.ini_to_dict(
+                    "/etc/ceph/%s.conf" % cluster_name
+                )
+                if "global" in raw_data:
+                    ret_val['detected_cluster_id'] = raw_data['global']\
+                        ['fsid']
+                    ret_val['cluster_attrs'] = {
+                        'fsid': raw_data['global']['fsid'],
+                        'name': 'ceph'
+                    }
 
             return ret_val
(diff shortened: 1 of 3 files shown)


shtripat:
@Tendrl/tendrl-core please review.
Verified the changes with mon and osd nodes both and it works as expected.

tendrl/node_agent/ini2json.py
```diff
@@ -0,0 +1,125 @@
+from ConfigParser import ConfigParser
```
nnDarshan: Can this file be moved to common utils ?

shtripat: Will do as part of another PR later as its referred in gluster-integration as well. Hope you are ok with that :)

nnDarshan: sure

tendrl/node_agent/discovery/sds/plugins/discover_ceph_sds_details.py
```diff
@@ -25,26 +26,15 @@ def discover_storage_system(self):
...
-            out, err = cmd.communicate()
-            if err:
-                LOG.error("Error getting cluster details")
-                return ret_val
-            if out:
-                details = json.loads(out)
-                ret_val['detected_cluster_id'] = details['fsid']
-                ret_val['cluster_attrs'] = {'fsid': details['fsid'],
-                                            'name': 'ceph'}
+            raw_data = ini2json.ini_to_dict("/etc/ceph/ceph.conf")
```
nthomas-redhat: This will not work if you have a custom cluster name. In that case the config file name is something like <cluster-name>.conf

shtripat: @nthomas-redhat so while sds detection there is no truth value from where we can get the cluster name. Without knowing the cluster name, how we think of loading the conf file here?

I feel underlying cluster could always be named `ceph` and console can have custom names known to it just for differentiating between multiple ceph clusters. Behind the scene always `fsid` is used for communicating with underlying cluster.

I certainly feel that looking for `/etc/ceph/*.conf` might be problematic at certain stage if ceph decides to have some other files in the same path later.

@Tendrl/tendrl-core comments?

nthomas-redhat:
> @nthomas-redhat so while sds detection there is no truth value from where we can get the cluster name. Without knowing the cluster name, how we think of loading the conf file here?
> 
> I feel underlying cluster could always be named ceph and console can have custom names known to it just for differentiating between multiple ceph clusters. Behind the scene always fsid is used for communicating with underlying cluster.
> 

That is not the case always. Think of an import cluster scenario where someone created a cluster with a custom name(you can do it through ceph-ansible)


> I certainly feel that looking for /etc/ceph/*.conf might be problematic at certain stage if ceph decides to have some other files in the same path later.
> 

Is there any other way go could get the fsid of the cluster from an OSD node?

> @Tendrl/tendrl-core comments?

shtripat:
@nthomas-redhat we may look for `ceph-osd` or `ceph-mon` process running status using `ps -ef` and get the cluster name from that. Using this name we can load the conf file.

The only thing is one by one we would have to run `ps -ef | grep ceph-mon` and `ps -ef | grep ceph-osd` to find this out.

What you feel, we should look for `/etc/ceph/*.conf` or try to get the cluster name using process running status?

nthomas-redhat:
> @nthomas-redhat we may look for ceph-osd or ceph-mon process running status using ps -ef and get the cluster name from that. Using this name we can load the conf file.
> 
> The only thing is one by one we would have to run ps -ef | grep ceph-mon and ps -ef | grep ceph-osd to find this out.
> 
> What you feel, we should look for /etc/ceph/*.conf or try to get the cluster name using process running status?

The ps option looks good to me

brainfunked:
@nthomas-redhat @shtripat ps output parsing must be avoided at all costs. It's extremely brittle and unreliable.

tendrl/node_agent/ini2json.py
```diff
@@ -0,0 +1,125 @@
...
+            return self.getint(section, option)
+        elif type is bool:
+            return self.getboolean(section, option)
+        else:
+            raise NotImplementedError()
+
+
+def ini_to_dict(ini_file_path):
+    cfg = StrictConfigParser()
+    f = open(ini_file_path)
```
shtripat: @brainfunked this piece of code is simply copied from gluster-integration. Later we plan to move this code to commons module and we can refine it that time.

shtripat:
Verified the code with mon and osd nodes. Works fine..

shtripat:
@Tendrl/tendrl-core please review

Issue #198, opened 2017-02-01
node-agent consumes huge amount of resources


dahorak:
Several recent versions of `tendrl-node-agent` constantly consumes huge amount of resources (CPU).

Check the the _%CPU_  column in the following (truncated) output from `top` (this is one time hit, but it's very similar for long period of time):
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                          
11192 root      20   0  357212  26528   5004 R  91.6  1.4   9:35.22 tendrl-node-age   
```

You can check it via following commands, which print the CPU and Memory usage of tendrl-node-agent each 5 second:
```
while true;
  do ps -p $(echo $(ps aux | grep [t]endrl-node-agent | awk '{print $2}') | sed 's/ /,/') -o %cpu,%mem,cmd -h;
  sleep 5;
done
```
For example (header added manually):
```
%CPU %MEM CMD
65.5  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.6  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.5  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.5  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.6  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.5  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
65.5  1.4 /usr/bin/python /usr/bin/tendrl-node-agent
```
When you stop the _tendrl-node-agent service_, it unfailingly lower the system load:
```
# uptime 
 15:35:21 up 33 min,  4 users,  load average: 1.20, 1.10, 0.98

# systemctl stop tendrl-node-agent
...
# uptime 
 15:38:33 up 36 min,  4 users,  load average: 0.05, 0.57, 0.79
```

This apply on the following versions of tendrl packages:
```
tendrl-commons-1.2-02_01_2017_18_29_12.noarch
tendrl-gluster-integration-1.2-02_01_2017_14_39_11.noarch
tendrl-node-agent-1.2-02_01_2017_14_21_11.noarch
```

dahorak:
It seems to be some regression in last two or three builds, because this is the graph of load of one of our hypervisors, where are running bunch of VMs with Tendrl.
[...]

PR #199, opened 2017-02-02
Span the hardware inventory collection over a period of time.

nnDarshan:
Currently the node inventory is synced every 3 seconds which is very aggressive, considering they change very rarely. So Its better to change the sync interval. Also instead of syncing everything after certain interval (say 5 or 10 mins) of time which causes a huge spike in CPU utilization every time this interval is reached, We can span the inventory collection over a period of time.

 In this patch I have introduced a gevent sleep of 10 secs after collecting every resource. So Its like every resource is updated every 3 minutes once.
tendrl-bug-id: https://github.com/Tendrl/node-agent/issues/198

### DIFF
--- a/tendrl/node_agent/node_sync/__init__.py
+++ b/tendrl/node_agent/node_sync/__init__.py
@@ -40,7 +40,12 @@ class NodeAgentSyncThread(sds_sync.StateSyncThread):
 
         while not self._complete.is_set():
             try:
-                gevent.sleep(3)
+                interval = 10
+                if tendrl_ns.first_node_inventory_sync:
+                    interval = 2
+                    tendrl_ns.first_node_inventory_sync = False
+                    
+                gevent.sleep(interval)
                 tags = []
                 # update node agent service details
                 LOG.info("node_sync, Updating Service data")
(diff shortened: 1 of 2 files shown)


nnDarshan:
@brainfunked @r0h4n @nthomas-redhat @shtripat Please review

anmolbabu:
None of these impact performance monitoring because that is in total done for us by collectd. But, in future when we add service status monitoring, we might only see a lag in the time of it being noticed and that very very minute and I think it is very much tolerable. And given that it is not delaying the first time update(Its only that interval b/w subsequent syncs has changed) specially of node context, the change looks good to me.

r0h4n:
@nnDarshan What is the cpu consumption with this PR?

tendrl/node_agent/node_sync/__init__.py
```diff
@@ -40,7 +40,12 @@ def _run(self):
 
         while not self._complete.is_set():
             try:
-                gevent.sleep(3)
+                interval = 10
+                if tendrl_ns.first_node_inventory_sync == True:
```
r0h4n: a simple "if tendrl_ns.first_node_inventory_sync" is enough

tendrl/node_agent/node_sync/__init__.py
```diff
@@ -49,20 +54,25 @@ def _run(self):
                     if s.running:
                         tags.append(TENDRL_SERVICE_TAGS[service.strip("@*")])
                     s.save()
+                    gevent.sleep(interval)
```
r0h4n: do not sleep for every iteration of this loop, put this after the loop

tendrl/node_agent/node_sync/__init__.py
nnDarshan: Done

tendrl/node_agent/node_sync/__init__.py
nnDarshan: Done

nnDarshan:
@r0h4n Earlier in the machine I tested CPU util was continuously 40-50%... now with this fix  and  https://github.com/Tendrl/commons/pull/151 its always below 1% and when we fetch inventory it spikes a little to around 5-10%

PR #200, opened 2017-02-02
Corrected sds detection plugin logic

shtripat:
In non-ceph nodes the file /etc/sysconfig/ceph does not exist and it would fail loading


shtripat:
@nnDarshan @r0h4n plz review

PR #204, opened 2017-02-08
Logging implementation for tendrl

GowthamShanmugam:
tendrl-spec: logging_implementation

Signed-off-by: root <root@dhcp42-75.lab.eng.blr.redhat.com>

### DIFF
--- a/etc/tendrl/node-agent/node-agent.conf.yaml.sample
+++ b/etc/tendrl/node-agent/node-agent.conf.yaml.sample
@@ -5,3 +5,4 @@ tendrl_ansible_exec_file: ~/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
+logging_socket_path: "/var/run/tendrl/message.sock."
--- a/tendrl/node_agent/message/handler.py
+++ b/tendrl/node_agent/message/handler.py
@@ -1,0 +1,63 @@
+import gevent.event
+import gevent.greenlet
+from gevent.server import StreamServer
+from gevent import socket
+from gevent.socket import error as socket_error
+from gevent.socket import timeout as socket_timeout
+from io import BlockingIOError
+import os
+import sys
+from tendrl.commons.message import Message
+from tendrl.node_agent.message.logger import Logger
+
+
+RECEIVE_DATA_SIZE = 4096
+
+
+class MessageHandler(gevent.greenlet.Greenlet):
+    def __init__(self):
+        super(MessageHandler, self).__init__()
+        self.server = StreamServer(
+            self.bind_unix_listener(),
+            self.read_socket
+        )
+
+    def read_socket(self, sock, address):
+        try:
+            self.data = sock.recv(RECEIVE_DATA_SIZE)
+            message = Message.from_json(self.data)
+            Logger(message)
+        except (socket_error, socket_timeout) as ex:
+            sys.stderr.write(
+                "Unable to read message from socket err.%s\n" % str(ex))
+        except (TypeError, ValueError, KeyError, AttributeError) as ex:
+            sys.stderr.write(
+                "Unable to log the message.%s\n" % self.data)
+
+    def _run(self):
+        try:
+            self.server.serve_forever()
+        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            sys.stderr.write(
+                "Unable to start socket .%s\n" % str(ex))
+
+    def stop(self):
+        socket_path = tendrl_ns.config.data['logging_socket_path']
+        self.sock.close()
+        if os.path.exists(socket_path):
+            os.remove(socket_path)
+        self.server.close()
+
+    def bind_unix_listener(self):
+        socket_path = tendrl_ns.config.data['logging_socket_path']
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        try:
+            if os.path.exists(socket_path):
+                os.remove(socket_path)
+            self.sock.setblocking(0)
+            self.sock.bind(socket_path)
+            self.sock.listen(50)
+        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            sys.stderr.write(
+                "Unable to create socket .%s\n" % str(ex))
+        return self.sock
--- a/tendrl/node_agent/message/logger.py
+++ b/tendrl/node_agent/message/logger.py
@@ -1,0 +1,103 @@
+import logging
+from tendrl.commons.message import Message
+
+LOG = logging.getLogger(__name__)
+
+
+class Logger(object):
+    logger_priorities = {"notice": "info",
+                         "info": "info",
+                         "error": "error",
+                         "debug": "debug",
+                         "warning": "warning",
+                         "critical": "critical"
+                         }
+
+    def __init__(self, message):
+        self.message = message
+        self.push_messages()
+        if self.message.request_id is not None:
+            """ If request_id is present then
+
+            it is considered as operation
+            """
+            self._logger(self.push_operation())
+        else:
+            self._logger(self.message.payload["message"])
+
+    def push_operation(self):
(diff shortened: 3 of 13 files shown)


GowthamShanmugam:
@r0h4n review please

GowthamShanmugam:
@nthomas-redhat @anmolbabu @brainfunked @anmolsachan @shtripat  please review this

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,3 +5,4 @@ tendrl_ansible_exec_file: $HOME/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
+logging_socket_path: "/tmp/node-agent/.node_agent.sock"
```
nnDarshan: I feel its better to have the socket file at "/var/run/tendrl/node-agent/.node_agent.sock" as /tmp is a bit vulnerable, user might clear /tmp directory.
@nthomas-redhat @shtripat please comment

tendrl/node_agent/message/handler.py
```diff
@@ -0,0 +1,53 @@
...
+            self.bind_unix_listener(),
+            self.read_socket
+        )
+
+    def read_socket(self, sock, address):
+        try:
+            data = sock.recv(RECEIVE_DATA_SIZE)
+            message = Message.from_json(data)
+            Logger(message)
+        except (TypeError, ValueError, KeyError,
```
nnDarshan: whats the point of these try except blocks ? if you are just raising the exception. This comment holds good for all the below try except blocls

tendrl/node_agent/message/handler.py
```diff
@@ -0,0 +1,53 @@
...
+                socket_error, socket_timeout) as ex:
+            raise ex
+
+    def _run(self):
+        try:
+            self.server.serve_forever()
+        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            raise ex
+
+    def stop(self):
```
nnDarshan: Does this take care of deleting the socket file ? If no, it has to be taken care. Else user will have to manually delete the socket file after every restart

tendrl/node_agent/message/handler.py
```diff
@@ -0,0 +1,53 @@
...
+from gevent.server import StreamServer
+from io import BlockingIOError
+import os
+from tendrl.commons.message import Message
+from tendrl.node_agent.message.logger import Logger
+
+
+RECEIVE_DATA_SIZE = 4096
+
+class MessageHandler(gevent.greenlet.Greenlet):
```
nnDarshan: If any issue occurs in this class, I feel it has to be written to stdout as the logger is not initialized yet and if not written to any place there is no way for user to know what went wrong.

etc/tendrl/node-agent/node-agent.conf.yaml.sample
GowthamShanmugam: @r0h4n please give your suggestion for this

tendrl/node_agent/message/handler.py
GowthamShanmugam: @nnDarshan if the user runs node_agent as service then how they will know where exception occurred, this problem also in event.py in commons, if socket fails i don't know what to do, if socket fails i am raising an exception in event.py but try catch is required for all the places where log message is created. we can handle error there but problem is for alerting anmol wants that exception in collecd.

tendrl/node_agent/message/handler.py
GowthamShanmugam: @nnDarshan should we print the exception?

tendrl/node_agent/message/handler.py
GowthamShanmugam: @nnDarshan i will do it

tendrl/node_agent/message/logger.py
```diff
@@ -0,0 +1,81 @@
...
+            else:
+                tendrl_ns.etcd_orm.client.write(
+                    ("nodes/%s/Messages/%s") % (
+                        self.message.node_id,
+                        self.message.message_id),
+                    Message.to_json(self.message)
+                )
+            # storing messages global wise
+            tendrl_ns.etcd_orm.client.write(
+                ("Messages/%s") % (
```
anmolbabu: Should this be lower case just as per normal convention "Messages" ==> "messages" ?

tendrl/node_agent/message/logger.py
```diff
@@ -0,0 +1,81 @@
...
+                    self.message.message_id),
+                Message.to_json(self.message)
```
anmolbabu: :+1: for dumping as json to etcd HERE

tendrl/node_agent/message/logger.py
GowthamShanmugam: @anmolbabu all the place we are using caps only, rohan suggested to me to like this

etc/tendrl/node-agent/node-agent.conf.yaml.sample
brainfunked: `/var/run/tendrl/message.sock`.

GowthamShanmugam:
@brainfunked i will change it

GowthamShanmugam:
@nthomas-redhat @brainfunked @shtripat @nnDarshan @anmolbabu @r0h4n @TimothyAsir @anmolsachan  please review

tendrl/node_agent/message/handler.py
nthomas-redhat: see the comment-->https://github.com/Tendrl/specifications/pull/94#issuecomment-279876382

tendrl/node_agent/message/handler.py
nthomas-redhat: Exceptions should always go into stderr along with configured logger

tendrl/node_agent/message/handler.py
```diff
@@ -0,0 +1,56 @@
...
+            print(traceback.format_exc())
+
+    def _run(self):
+        try:
+            self.server.serve_forever()
+        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            print(traceback.format_exc())
+
+    def stop(self):
+        self.sock.close()
```
nthomas-redhat: Also handle the exceptions in this function

tendrl/node_agent/message/logger.py
```diff
@@ -0,0 +1,81 @@
+import logging
+from tendrl.commons.message import Message
+
+
+LOG = logging.getLogger(__name__)
+
+
+class Logger(object):
+    logger_priorities = {"notice": "info",
+                         "info": "info",
```
nthomas-redhat: Looks like 'info' and 'notice' are same. why you need both?
Is it done for compatebility with common logging?

GowthamShanmugam: @nthomas-redhat alert are always comes as notice, anmol can pick the alert messages from etcd which is in notice type. 

We are storing all the messages in  etcd under /Messages. Anmol can tack alert from etcd using which have type notice.  But in log there is no notice function so i logging notice as info

GowthamShanmugam:
@nthomas-redhat @shtripat @nnDarshan @r0h4n @brainfunked @anmolbabu @TimothyAsir please review

tendrl/node_agent/message/handler.py
```diff
@@ -0,0 +1,63 @@
...
+from gevent.socket import error as socket_error
+from gevent.socket import timeout as socket_timeout
+from io import BlockingIOError
+import os
+import sys
+from tendrl.commons.message import Message
+from tendrl.node_agent.message.logger import Logger
+
+
+RECEIVE_DATA_SIZE = 4096
```
brainfunked: What's the rationale behind this specific size?

GowthamShanmugam: @brainfunked When reading from a socket it's always required to specify reading byte size, it means never return until read x bytes from the socket.

Issue #209, opened 2017-02-14
Mark import job as finished only after the imported cluster appers in etcd


nnDarshan:
We have to mark the job for import cluster as finished only after imported cluster appears in the etcd. Else this will lead to: https://github.com/Tendrl/api/issues/67

PR #210, opened 2017-02-14
Mark the import cluster job status correctly

nnDarshan:
We have to mark the job for import cluster as finished
only after imported cluster appears in the etcd. Else
it might cause inconsistency in API layer.

tendrl-bug-id: Tendrl/node-agent#209

### DIFF
--- a/tendrl/node_agent/flows/import_cluster/__init__.py
+++ b/tendrl/node_agent/flows/import_cluster/__init__.py
@@ -1,3 +1,5 @@
+import etcd
+import gevent
 import json
 import uuid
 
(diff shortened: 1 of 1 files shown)


nnDarshan:
This fixes the issue: https://github.com/Tendrl/api/issues/67

nnDarshan:
@shtripat @nthomas-redhat @r0h4n Please review

tendrl/node_agent/flows/import_cluster/__init__.py
```diff
@@ -42,3 +44,16 @@ def run(self):
...
+        # function is returned. This might lead to inconsistancy in the API
+        # functionality. The below loop waits for the cluster details
+        # to be updated in etcd.
+        while True:
+            gevent.sleep(2)
+            try:
+                tendrl_ns.etcd_orm.client.read("/clusters/%s" % integration_id)
+                break
+            except etcd.EtcdKeyNotFound:
+                continue
```
shtripat: How do we mark a task as failed if some exception etc happens while execution of the flows? I feel the same should be handled better way in commons job handler. Can you create an issue and send a PR for the same?

nnDarshan: The common job handler expects a flow_execution_failed exception and marks job as failed if this exception is raised in the flow. In this case, since we have stop gap gluster and ceph import flows they are not raised. When they are implemented properly they will be raising the exception when some issue occures

shtripat: Ok. Then its fine.

PR #211, opened 2017-02-17
Logging implementation for tendrl

GowthamShanmugam:
tendrl-spec: logging_implementation

### DIFF
--- a/etc/tendrl/node-agent/node-agent.conf.yaml.sample
+++ b/etc/tendrl/node-agent/node-agent.conf.yaml.sample
@@ -5,3 +5,4 @@ tendrl_ansible_exec_file: $HOME/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
+logging_socket_path: "/var/run/tendrl/.message.sock."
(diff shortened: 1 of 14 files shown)


GowthamShanmugam:
@brainfunked @shtripat @r0h4n @nthomas-redhat @nnDarshan @anmolbabu please review

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,3 +5,4 @@ tendrl_ansible_exec_file: $HOME/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
+logging_socket_path: "/var/run/tendrl/message.sock."
```
shtripat: may be keep the socket file name as `/var/run/tendrl/.message.sock`

tendrl/node_agent/objects/definition/master.py
```diff
@@ -285,6 +285,126 @@
...
+          help: "parent id"
+          type: String
+        cluster_id:
+          help: "cluster id"
+          type: String
+        caller:
+          help: "Called details"
+          type: Dict
+      enabled: true
+      list: $request_id
```
shtripat: Is this correct path?

etc/tendrl/node-agent/node-agent.conf.yaml.sample
GowthamShanmugam: @shtripat i will change it

tendrl/node_agent/objects/definition/master.py
GowthamShanmugam: @shtripat job updates stored only using request id, then this should br the key right?

PR #214, opened 2017-02-17
Added creation of socket directory for events

shtripat:

### DIFF
--- a/etc/tendrl/node-agent/node-agent.conf.yaml.sample
+++ b/etc/tendrl/node-agent/node-agent.conf.yaml.sample
@@ -5,4 +5,4 @@ tendrl_ansible_exec_file: $HOME/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
-logging_socket_path: "/var/run/tendrl/.message.sock."
+logging_socket_path: /var/run/.tendrl.message.sock


shtripat:
@GowthamShanmugam the directory creation for /var/run/tendrl was missing

shtripat:
@Tendrl/tendrl-core there is an issue with socket file creation. The directory `/var/run/tendrl` doesnt persist across reboots of the machine. So either we need some mechanism to create this directory everytime `node-agent` is started or create the socket file with path `/var/run/.tendrl.message.sock`.

I feel second option is easier to implement and reliable as well. Suggestions?

GowthamShanmugam:
@shtripat i have added a code in tendrl-node-agent to create a runtime directory RuntimeDirectory=tendrl, but it works when we run node-agent-as service, when we run dircctly using command tendrl-node-agent then directory is not created

GowthamShanmugam:
@shtripat i accept you suggestion of creating directory ... @brainfunked you suggestion please

nthomas-redhat:
Don't do it manually. Let the systemd do it for you. Place a configuration file for tendrl in /usr/lib/tmpfiles.d. During the boot process, systemd-tmpfiles populates /run based on that confguration

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,4 +5,4 @@ tendrl_ansible_exec_file: $HOME/.tendrl/node-agent/ansible_exec
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
-logging_socket_path: "/var/run/tendrl/.message.sock."
+logging_socket_path: /var/run/.tendrl.message.sock
```
nthomas-redhat: Please have a look at the comment https://github.com/Tendrl/node-agent/pull/214#issuecomment-280911086

shtripat:
As per comments from @nthomas-redhat we would keep the path as `/var/run/tendrl/.message.sock`. Abandoning this change.
@GowthamShanmugam plz send a PR for add required changes in systemd file.

GowthamShanmugam:
@shtripat i will sent it

etc/tendrl/node-agent/node-agent.conf.yaml.sample
GowthamShanmugam: @nthomas-redhat i got your point i will update this patch as your per you suggestion

brainfunked:
This needs to be corrected in a couple of ways:

1. The socket is a public interface to Tendrl functionality and should not be hidden.
2. The socket should be managed by systemd.

@GowthamShanmugam @nthomas-redhat @TimothyAsir

PR #216, opened 2017-02-18
Adding permission for tendrl group to access message socket file

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#223

### DIFF
--- a/tendrl/node_agent/message/handler.py
+++ b/tendrl/node_agent/message/handler.py
@@ -6,6 +6,7 @@ from gevent.socket import error as socket_error
 from gevent.socket import timeout as socket_timeout
 from io import BlockingIOError
 import os
+import grp
 import sys
 from tendrl.commons.message import Message
 from tendrl.node_agent.message.logger import Logger
@@ -57,7 +58,13 @@ class MessageHandler(gevent.greenlet.Greenlet):
             self.sock.setblocking(0)
             self.sock.bind(socket_path)
             self.sock.listen(50)
-        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            gid = grp.getgrnam("tendrl").gr_gid
+            os.chown(socket_path, -1, gid)
+            os.chmod(socket_path, 0770)
+        except(TypeError, BlockingIOError, socket_error, ValueError) as ex:
             sys.stderr.write(
                 "Unable to create socket .%s\n" % str(ex))
+        except(KeyError) as ex:
+            sys.stderr.write(
+                "User not found: tendrl-user .err:%s\n" % str(ex))
         return self.sock


GowthamShanmugam:
@brainfunked @nthomas-redhat @anmolbabu @shtripat @r0h4n please review

tendrl/node_agent/message/handler.py
```diff
@@ -57,7 +58,13 @@ def bind_unix_listener(self):
             self.sock.setblocking(0)
             self.sock.bind(socket_path)
             self.sock.listen(50)
-        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            gid = pwd.getpwnam("tendrl-user").pw_gid
```
nthomas-redhat: When 'tendrl-user' is created? and who creates it
Is it a good idea to assign permissions for all the users in a group rather than doing it on per user basis? @brainfunked , @r0h4n

anmolbabu: @nthomas-redhat This user is required for collectd to run a custom plugin on new notification because of its(collectd's) limitation that it can't run a plugin on notification(threshold breach detected) as root.
Also, please check the my comment above...

GowthamShanmugam: @nthomas-redhat here i am giving permission for that user and entire group is wrong, i have to give permission to that user only like os.chown(socket_path, uid, -1), -1 is used to leave one of the ids unchanged.

tendrl/node_agent/message/handler.py
```diff
@@ -57,7 +58,12 @@ def bind_unix_listener(self):
             self.sock.setblocking(0)
             self.sock.bind(socket_path)
             self.sock.listen(50)
-        except (TypeError, BlockingIOError, socket_error, ValueError) as ex:
+            uid = pwd.getpwnam("tendrl-user").pw_uid
```
brainfunked: System uid should be just `tendrl`.

tendrl/node_agent/message/handler.py
```diff
@@ -57,7 +58,12 @@ def bind_unix_listener(self):
...
+            os.chown(socket_path, uid, -1)
```
brainfunked: The socket permissions should be `770`, owned by `root` and group ownership set to `tendrl`.

tendrl/node_agent/message/handler.py
GowthamShanmugam: @brainfunked ok i will change it

tendrl/node_agent/message/handler.py
GowthamShanmugam: @brainfunked i will change it

GowthamShanmugam:
@shtripat ok

GowthamShanmugam:
@brainfunked i under stand your point @anmolbabu your suggestion please

anmolbabu:
@GowthamShanmugam I remember we tried out manually whatever Mrugest suggested and it had worked. But I remember seeing @TimothyAsir having a PR to do the same but in another(systemd) way so please check with him once

GowthamShanmugam:
@anmolbabu Tim have a patch for create directory for socket ( /var/run/tendrl ) he is giving a permission like 660 so it will accessible by root and group. And he is going to create tendrl-user also please. i feel this patch is not required.

anmolbabu:
@GowthamShanmugam yes. Probably you can abandon this ...

PR #221, opened 2017-02-20
Logging changes according to new logging framework 

anmolsachan:
tendrl-spec: Tendrl/specifications#127
reference-spec: Tendrl/specifications#94
reference-code-pr: Tendrl/commons#161

Signed-off-by: anmolsachan asachan@redhat.com

### DIFF
--- a/tendrl/node_agent/discovery/platform/manager.py
+++ b/tendrl/node_agent/discovery/platform/manager.py
@@ -1,12 +1,12 @@
 import importlib
 import inspect
-import logging
 import os
+import traceback
+
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
 
 from tendrl.node_agent.discovery.platform import base
-
-
-LOG = logging.getLogger(__name__)
 
 
 class PlatformManager(object):
@@ -30,8 +30,16 @@ class PlatformManager(object):
                 for name, cls in clsmembers:
                     exec("from %s import %s" % (plugin_name, name))
         except (SyntaxError, ValueError, ImportError) as ex:
-            LOG.error('Failed to load the platform plugins. Error %s' %
-                      ex, exc_info=True)
+            Event(
+                Message(
+                    priority="error",
+                    publisher=tendrl_ns.publisher_id,
+                    payload={"message": 'Failed to load the platform plugins. '
+                                        'Error %s %s' %
+                                        (ex, traceback.format_exc())
+                             }
+                )
+            )
             raise ex
         return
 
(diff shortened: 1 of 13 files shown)


anmolsachan:
@GowthamShanmugam @shtripat please review

tendrl/node_agent/discovery/platform/manager.py
```diff
@@ -30,8 +30,16 @@ def load_plugins(self):
...
+            Event(
+                Message(
+                    priority="error",
+                    publisher=tendrl_ns.publisher_id,
+                    payload={"message": 'Failed to load the platform plugins. '
+                                        'Error %s %s' %
+                                        (ex, traceback.format_exc())
+                             }
+                )
+            )
```
GowthamShanmugam: For this you can use exception message class but it is not merged if it is merged then change as per that

tendrl/node_agent/node_sync/disk_sync.py
```diff
@@ -79,9 +78,21 @@ def get_node_disks():
...
                             rv['used_disks_id'].append(disk['disk_id'])
         else:
-            LOG.error(err)
+            Event(
+                Message(
+                    priority="error",
+                    publisher=tendrl_ns.publisher_id,
+                    payload={"message": err}
+                )
+            )
```
GowthamShanmugam: Make sure this is 'err' is string, other wise it gives  serialization problem, If Exception message class came then you can use exception as it is. But now you have to use string only.

tendrl/node_agent/node_sync/disk_sync.py
```diff
@@ -79,9 +78,21 @@ def get_node_disks():
...
     else:
-        LOG.error(err)
+        Event(
+            Message(
+                priority="error",
+                publisher=tendrl_ns.publisher_id,
+                payload={"message": err}
+            )
+        )
```
GowthamShanmugam: In exception class you can use exception object, but now you can't use exception object it gives serialization problem so make sure this "ex" is string or not

tendrl/node_agent/discovery/platform/manager.py
anmolsachan: Yeah. I will change this later

tendrl/node_agent/node_sync/disk_sync.py
anmolsachan: Made the changes

tendrl/node_agent/node_sync/disk_sync.py
anmolsachan: Made the changes

PR #236, opened 2017-03-03
improve configuration of setuptools, tox and travis

mbukatov:
This pull request updates tendrl-node-agent based on improvements added into tendrl-commons recently.

tendrl-bug-id: Tendrl/node-agent#235


setup.py
```diff
@@ -118,7 +108,14 @@ def run(self):
...
     license="LGPL-2.1+",
     zip_safe=False,
-    install_requires=install_requires,
+    install_requires=[
+        "PyYAML",
+        "gevent>=1.0",
+        "namespaces",
+        "python-etcd",
+        "six",
+        "tendrl-commons",
```
r0h4n: This would require "tendrl-commons" to be available via pip install (pypi)?

mbukatov: If we were installing node-agent via pip install, then yes. But since we can't do a full installation this way anyway (system default configuration can't be installed into /etc, dependencies on non python components related to ceph or gluster can't be listed, ...) it's not a problem. This is related to my proposal I send to tendrl-devel some time ago, which suggested to simplify packaging approach by (among other things) stating that end user should not install the component via pip install, but via system (rpm or deb) packages only.

My idea is to have a primary source of dependencies in a single place, in install_requires of setup.py even when we can't install the component solely via pip install, because it would help us to clearly track list of components, avoiding keeping unnecessary ones or missing others.

For development or CI environment, one can install the component via requirements.txt files, which are reusing dep list from setup.py, but it also specifies where to get tendrl-commons, so that we don't have to publish tendrl-commons on PyPI for this to work:

```
$ cat requirements.master.txt
# use latest version from master branch on github
git+https://github.com/Tendrl/commons.git@master
-e .
```

This way, we can install the component for dev or CI purposes like this:

```
$ pip install -r requirements.master.txt
```

r0h4n:
@mbukatov Please resolve merge conflicts

mbukatov:
@r0h4n I rebased branch

PR #255, opened 2017-03-08
Reading jobs based on new job class

GowthamShanmugam:
Signed-off-by: root <root@dhcp42-230.lab.eng.blr.redhat.com>


GowthamShanmugam:
@r0h4n @shtripat @brainfunked @nthomas-redhat

PR #256, opened 2017-03-08
fix for network interface in node_sync

GowthamShanmugam:
Signed-off-by: root <root@dhcp41-178.lab.eng.blr.redhat.com>


GowthamShanmugam:
@r0h4n

GowthamShanmugam:
@r0h4n please merge this

tendrl/node_agent/node_sync/__init__.py
```diff
@@ -163,7 +163,7 @@ def _run(self):
                             # it will delete subnet dir when it is empty
                             # if one entry present then deletion never happen
                             NS.etcd_orm.client.delete(network.key, dir=True)
```
r0h4n: Why are you trying to delete subnet dir on every sync?

GowthamShanmugam: @r0h4n all the network interface are stored under subnet, if all interface are down in particular subnet then subnet path contains empty  and and empty subnet path always present etcd. so it will delete subnet when subnet path completely empty

PR #258, opened 2017-03-08
Setup ssh keys on storage nodes for provisioning

nthomas-redhat:

### DIFF
--- a/tendrl/node_agent/provisioner/ceph/plugins/ceph_installer.py
+++ b/tendrl/node_agent/provisioner/ceph/plugins/ceph_installer.py
@@ -1,6 +1,7 @@
 
 import json
 import logging
+import os
 import urllib3
 
 from tendrl.node_agent.provisioner.ceph.provisioner_base import\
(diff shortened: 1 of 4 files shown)


tendrl/node_agent/provisioner/ceph/plugins/ceph_installer.py
```diff
@@ -252,3 +253,26 @@ def task_status(self, task_id):
...
+    def setup(self, provisioner_node_fqdn):
+        url = 'http://%s:%s/setup/' % (
+            provisioner_node, self._CEPH_INSTALLER_API_PORT)
+        resp = self.http.request(
+            self._MGET,
+            url)
+        if resp.status == 200:
+            resp_data = resp.data.decode("utf-8")
+            try:
+                f = open("/tmp/setup.sh", "w")
```
shtripat: We should delete this file after successful execution below.

shtripat: Also not sure if `/tmp/setup.sh` is the path we should use. I feel we should have a tendrl specific directory where we create these tendrl specific temporary files and delete them after using it.

PR #278, opened 2017-03-14
Fix for messages in read message function

GowthamShanmugam:
tendrl-bug-id:Tendrl/node-agent#279

Read message function takes only two arguments as
parameter but gevent is passing 3 parameters for read function.
Third is args.


GowthamShanmugam:

tendrl/node_agent/objects/cluster_message/__init__.py
```diff
@@ -10,6 +10,9 @@ def __init__(self, **cluster_message):
         self.value = 'clusters/%s/Messages/%s'
         self._etcd_cls = _ClusterMessageEtcd
 
+    def load_definition(self):
```
shtripat: Is it mandatory to have `load_definitions()` with empty implementation?

GowthamShanmugam: @shtripat with out this message class gives error

shtripat: ok

r0h4n: overloading load_definitions is a way to bypass the definitions system (i.e you dont have to define the object/flow/atom in the definitions file)

PR #280, opened 2017-03-15
fix in cluster node context for import cluster

GowthamShanmugam:


GowthamShanmugam:
@brainfunked @shtripat please review

PR #293, opened 2017-03-16
Stored job_updates only under /Messages/job

GowthamShanmugam:
The job_updates are stored under /Messages/job and
/Message/event also now this will fix job_updates are
stored only in /Messages/job.


GowthamShanmugam:

PR #294, opened 2017-03-16
Added gluster provisioner

shtripat:


shtripat:

shtripat:
Verified the changes with both gluster and ceph cluster creation flows. Works as expected.

etc/tendrl/node-agent/node-agent-dev.conf.yaml
```diff
@@ -4,3 +4,8 @@ etcd_connection: 0.0.0.0
 log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: pip
+logging_socket_path: /var/run/tendrl/message.sock
+ceph_provisioner: CephInstallerPlugin
```
r0h4n: This needs to be hardcoded in tendrl.objects.definitions.master (.yml) , There is no reason strong enough to let end users understand theses internals and handle them via config

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,5 +5,7 @@ log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
+ceph_provisioner: CephInstallerPlugin
+gluster_provisioner: GdeployPlugin
```
r0h4n: Same as above file

tendrl/node_agent/manager/__init__.py
```diff
@@ -67,18 +72,19 @@ def main():
...
     NS.compiled_definitions.save()
     NS.node_context.save()
     NS.tendrl_context.save()
     NS.node_agent.definitions.save()
     NS.node_agent.config.save()
     NS.message_handler_thread = MessageHandler()
-
     NS.publisher_id = "node_agent"
 
+    NS.ceph_provisioner = CephProvisioningManager(NS.config.data['ceph_provisioner'])
```
r0h4n: Get NS.config.data['ceph_provisioner'] from NS.tendrl.definitions.get_parsed_defs()["namespace.tendrl"]['ceph_provisioner']

r0h4n:
Also, The gdeploy provisioning code should go in commons , it is used by node-agent and gluster-integraton both

https://github.com/Tendrl/gluster-integration/pull/186/files#diff-27a10665b8a615a0f97f6796f532f27eR30

etc/tendrl/node-agent/node-agent-dev.conf.yaml
shtripat: OK. will move to yaml file

etc/tendrl/node-agent/node-agent.conf.yaml.sample
shtripat: Sure

tendrl/node_agent/manager/__init__.py
shtripat: ok

shtripat:
Regarding moving the code to commons, will discuss once with @nnDarshan as well and would do the needful.

etc/create_gluster_cluster_sample_job.py
```diff
@@ -0,0 +1,41 @@
...
+        "sds_type": "glusterfs",
+        "sds_version": "3.2.1",
+        "name": "MyCluster",
+        "fsid": "140cd3d5-58e4-4935-a954-d946ceff371d",
+        "public_network": "10.70.40.0/22",
+        "cluster_network": "10.70.40.0/22",
+        "node_configuration": {
+            "ab003bbc-00cd-4b74-b236-f19c2c33b96b": {
+                "role": "glusterfs/node",
+                "provisioning_ip": "10.70.42.183",
```
nthomas-redhat: what is provisioning_ip here?
In gluster's case its mostly the management ip address.
@brainfunked , comments please

etc/create_gluster_cluster_sample_job.py
```diff
@@ -0,0 +1,41 @@
...
+    "created_at": "2017-03-09T14:15:14Z",
+    "username": "admin",
+    "parameters": {
+        "Node[]": ["81720b6c-6732-49dd-ad32-15845f199c79",
+                   "ab003bbc-00cd-4b74-b236-f19c2c33b96b"],
+        "sds_type": "glusterfs",
+        "sds_version": "3.2.1",
+        "name": "MyCluster",
+        "fsid": "140cd3d5-58e4-4935-a954-d946ceff371d",
+        "public_network": "10.70.40.0/22",
```
nthomas-redhat: public/cluster network looks like more of a ceph related term?

etc/create_gluster_cluster_sample_job.py
```diff
@@ -0,0 +1,41 @@
...
+        "node_configuration": {
+            "ab003bbc-00cd-4b74-b236-f19c2c33b96b": {
+                "role": "glusterfs/node",
+                "provisioning_ip": "10.70.42.183",
+            },
+        },
+        "TendrlContext.integration_id": "89604c6b-2eff-4221-96b4-e41319240240"
+    },
+    "node_ids": [],
+    "tags": ["provisioner/gluster"]
```
nthomas-redhat: In gluster's case this is one of the node in the cluster right?
Who assigns this role to a specific node? can you brief the workflow please?

etc/create_gluster_cluster_sample_job.py
shtripat: This is actually not used in case of gluster, so I feel can remove from sample program.

etc/create_gluster_cluster_sample_job.py
shtripat: My understanding of this field is IP with which the node would be provisioned.
@brainfunked ??

etc/create_gluster_cluster_sample_job.py
shtripat: >> In gluster's case this is one of the node in the cluster right?
yes

>> Who assigns this role to a specific node? can you brief the workflow please?
User needs to add this tag for one of the nodes at the moment. The work flow here would be like
- Identify a node as `provisioner/gluster` with tag entry and submit the create cluster job to this provisioner node.
- Provisioner node's ssh key is generated
- Provisioner node's ssh key is copied to other nodes by running individual flows on other nodes with ssh-key as one of the flow parameters
- gdeploy wrapper invoked to install gluster bits on the storage nodes
- gdeploy wrapper configures gluster service (enabled glusterd service and starts it)
- gdeploy wrapper configure firewalld details for gluster
- gdeploy creates gluster cluster using the nodes

Issue #297, opened 2017-03-16
monitor secret stored in code


mkudlej:
There is monitor secret stored in code, see https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/provisioner/ceph/plugins/ceph_installer.py#L139
```
...
        data = {
            "calamari": False,
            "host": host,
            "address": ip_address,
            "fsid": cluster_id,
            "monitor_secret": "AQA7P8dWAAAAABAAH/tbiZQn/40Z8pr959UmEA==",
            "cluster_name": cluster_name,
            "cluster_network": cluster_network,
            "public_network": public_network,
            "redhat_storage": False,
            "verbose": False
        }
        if monitors is not None:
          data.update({"monitors": monitors})
...
```

I think it should be replaced by code which inputs this secret from configuration, parameter or it is generated.

shtripat:
This can be closed as PR is merged already

PR #300, opened 2017-03-17
Added logic to generate monitor secret

shtripat:
tendrl-bug-id: Tendrl/node-agent#297


shtripat:

tendrl/node_agent/provisioner/ceph/utils.py
```diff
@@ -0,0 +1,16 @@
+import os
+import struct
+import time
+import base64
+
+
+def generate_auth_key():
```
r0h4n: please add comment explaining what type of key is required to be generated and how it is generated

shtripat: Sure. will add comments.

Issue #310, opened 2017-03-20
After copying authorize_key also ssh password prompt is asking yes/no


GowthamShanmugam:
During cluster creation flow the provisioning node asking yes/no prompt for each remote node's ssh connection. Because of this problem cluster creation flow can't proceed further.

nthomas-redhat:
Closing the issue as the PR is already merged

PR #312, opened 2017-03-20
Set environment variable for avoiding ssh connection yes/no prompt

GowthamShanmugam:
tendrl-bug-id:Tendrl/node-agent#310


GowthamShanmugam:
@r0h4n @shtripat

Issue #251, opened 2017-03-08
TTL on the message directories


brainfunked:
The message directories used for logging messages in etcd need to have a configurable TTL. Given the nature of etcd, it is necessary to expire old data periodically. The default interval could be something like 2 days. However, the interval needs to be administrator configurable, depending upon the memory available and the size of the cluster etc.

GowthamShanmugam:
i am working on this

shtripat:
One thought here.

time to live is I feel the simplest way to handle data load on central store.
But actual live scenario it should more be policy driven. For example, user should be able to decide what to do we audit log, operational flow data etc which are older than certain time. Few options could be as below

- Move the records to a different etcd instance which is archive of such old details. This could used later if required for audit purpose. This archive etcd instance maintains the same structure as of tendrl central store.
- Move the older details in the form of json dumps to external (configured) location. Optionally the the json dump files could be archived in the form of TAR files with date time stamp attached to name. This could later in imported in an etcd instance for reference

@Tendrl/tendrl-core comments? I feel this should be our ultimate goal.

GowthamShanmugam:
ttl is deleting only entries in the path but directories are not deleted, it show empty directory

GowthamShanmugam:
PR is under review

GowthamShanmugam:
Patch for assigning TTL to message and job_updates are done.

nthomas-redhat:
closing this as fix is merged

PR #314, opened 2017-03-22
Message retention period

GowthamShanmugam:
tendrl-bug-id:Tendrl/node-agent#251

### DIFF
--- a/etc/tendrl/node-agent/node-agent.conf.yaml.sample
+++ b/etc/tendrl/node-agent/node-agent.conf.yaml.sample
@@ -5,6 +5,6 @@ log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
-message_retention_time: 172800 # default 2 days
+message_retention_time: 172800 # default 2 days(2*24*60*60)
 tags:
 - tendrl/node
--- a/etc/tendrl/node-agent/node-agent.conf.yaml.sample
+++ b/etc/tendrl/node-agent/node-agent.conf.yaml.sample
@@ -5,5 +5,6 @@ log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
+message_retention_time: 172800 # default 2 days
 tags:
 - tendrl/node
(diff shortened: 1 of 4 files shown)


GowthamShanmugam:
@nthomas-redhat @shtripat @r0h4n please review

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,5 +5,6 @@ log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
+message_retention_time: 172800 # default 2 days
```
GowthamShanmugam: 2 days in seconds

shtripat: I would suggest to add `2*24*60*60` in comments

GowthamShanmugam: @shtripat ok

tendrl/node_agent/message/logger.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, message):
     def push_operation(self):
         NS.etcd_orm.client.write(
             "/Messages/jobs/%s" % self.message.job_id,
+            '',
+            dir=True,
+            ttl=NS.config.data['message_retention_time'])
+        NS.etcd_orm.client.write(
+            "/Messages/jobs/%s" % self.message.job_id,
```
GowthamShanmugam: because of sequence id ttl is not work here, that why i am writing base dir first with ttl and then i am writing content with sequence id

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -5,5 +5,6 @@ log_cfg_path: /etc/tendrl/node-agent/node-agent_logging.yaml
 log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
+message_retention_time: 172800 # default 2 days(2*24*60*60)
```
r0h4n: add the same "message_retention_time" in https://github.com/Tendrl/node-agent/blob/develop/etc/tendrl/node-agent/node-agent-dev.conf.yaml

GowthamShanmugam:
@r0h4n done

tendrl/node_agent/message/logger.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, message):
     def push_operation(self):
         NS.etcd_orm.client.write(
             "/Messages/jobs/%s" % self.message.job_id,
+            '',
```
r0h4n: Why are you passing an empty string as value? It is not required https://github.com/jplana/python-etcd/blob/0.4.5/src/etcd/client.py#L482

PR #330, opened 2017-03-27
Fixed startup issue due to latest changes

shtripat:
tendrl-bug-id: Tendrl/commons#307
Signe-off-by: Shubhendu <shtripat@redhat.com>

### DIFF
--- a/tendrl/node_agent/objects/definition/node_agent.yaml
+++ b/tendrl/node_agent/objects/definition/node_agent.yaml
@@ -13,8 +13,8 @@ namespace.node_agent:
     Config:
         enabled: True
         help: "Config"
-        value: _tendrl/config/data
-        list: _tendrl/config/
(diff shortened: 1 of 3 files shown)


tendrl/node_agent/objects/definition/node_agent.yaml
```diff
@@ -13,8 +13,8 @@ namespace.node_agent:
     Config:
         enabled: True
         help: "Config"
-        value: _tendrl/config/data
-        list: _tendrl/config/
+        value: _NS/tendrl/config/node-agent/data
```
r0h4n: value: "_NS/node_agent/config/data"
list: "_NS/node_agent/config"

shtripat: ok

PR #346, opened 2017-03-28
Fix socket start issue

TimothyAsirJeyasing:
Socket service will be automatically started/stopped once the
tendrl-node-agent service is started or stopped.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,7 +3,7 @@ include AUTHORS
 include README.rst
 include ChangeLog
 include tendrl-node-agent.service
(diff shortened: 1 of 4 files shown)


TimothyAsirJeyasing:
@r0h4n , @nthomas-redhat , @anmolbabu , @GowthamShanmugam , please review

TimothyAsirJeyasing:
The socket is already managed by systemd includes listening and binding. Currently there are few issues in the handler.py which is in tendrl/node_agent/message directory. It should only use the existing socket and should not create its own.

PR #359, opened 2017-03-30
Use tags defined in master.yml for updating tags

nnDarshan:
This patch removes the locally hardcoded tag details
and instead uses the globally defined tag details in
the master.yml


nnDarshan:

Issue #373, opened 2017-04-10
tendrl-node-agent fails to start after reboot


ltrilety:
Sometimes when machine is rebooted `tendrl-node-agent` fails to start.

```
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: Finding flows in namespace.node_agent.flowsSetup Tendrl definitions (.yml) for namespace.node_agentValidating registered (.py) objects in namespace.node_agent.objectsValidating defined (.yml) objects in namespace.node_agent.objectsSetup Config for namespace.node_agentSetup Etcd Orm for namespace.node_agentTraceback (most recent call last):
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: File "/usr/bin/tendrl-node-agent", line 9, in <module>
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: load_entry_point('tendrl-node-agent==1.2.2', 'console_scripts', 'tendrl-node-agent')()
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: File "/usr/lib/python2.7/site-packages/tendrl/node_agent/manager/__init__.py", line 40, in main
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: node_agent.NodeAgentNS()
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: File "/usr/lib/python2.7/site-packages/tendrl/node_agent/__init__.py", line 14, in __init__
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: super(NodeAgentNS, self).__init__(ns_name, ns_src)
Apr 10 12:47:24 ltrilety-usm1-gl4.usmqe.lab.eng.brq.redhat.com tendrl-node-agent[1316]: File "/usr/lib/python2.7/site-packages/tendrl/commons/__init__.py", line 63, in __init__
[...]

PR #393, opened 2017-04-19
Fix service start issue due to network unavailable

TimothyAsirJeyasing:
Update the script to just guarantees that the network service
has been started, Added Order after network target also for
compatibility with older systems


TimothyAsirJeyasing:
Tendrl bug url: https://github.com/Tendrl/node-agent/issues/373

@r0h4n , @nthomas-redhat , @shtripat , Please review

PR #402, opened 2017-04-26
WIP - Added flow for expand cluster

shtripat:

### DIFF
--- a/tendrl/node_agent/flows/expand_cluster/__init__.py
+++ b/tendrl/node_agent/flows/expand_cluster/__init__.py
@@ -1,0 +1,65 @@
+from tendrl.commons import flows
+from tendrl.commons.event import Event
+from tendrl.commons.flows.create_cluster import \
+    utils as create_cluster_utils
+from tendrl.commons.flows.exceptions import FlowExecutionFailedError
+from tendrl.commons.message import Message
+from tendrl.node_agent.flows.expand_cluster import ceph_help
+
+
+class ExpandCluster(flows.BaseFlow):
+    def run(self):
+        integration_id = self.parameters['TendrlContext.integration_id']
+        if integration_id is None:
+            raise FlowExecutionFailedError("TendrlContext.integration_id cannot be empty")
+
+        supported_sds = NS.compiled_definitions.get_parsed_defs()['namespace.tendrl']['supported_sds']
+        sds_name = self.parameters["TendrlContext.sds_name"]
+        if sds_name not in supported_sds:
+            raise FlowExecutionFailedError("SDS (%s) not supported" % sds_name)
+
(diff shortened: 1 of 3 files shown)


shtripat:
Verified the OSD node addition and works fine.
MON node addition still has some glitches. Need to correct and verify.

shtripat:
There is a ceph BZ created https://bugzilla.redhat.com/show_bug.cgi?id=1446396 for MON addition issue.
Had tested the addition of OSD nodes earlier and worked as required.

We can merge the PR with a note that MON addition need to be re-looked once above BZ is fixed.

@Tendrl/tendrl-core comments?

r0h4n:
The BZ is closed https://bugzilla.redhat.com/show_bug.cgi?id=1446396

The last comment on the BZ advises to use a newer version of the ceph-installer

shtripat:
Yes @r0h4n will try to verify the scenario with latest ceph-installer once and update here.

r0h4n:
these flows have to be in the root namespace which is "NS.tendrl" along with NS.tendrl.flows.CreateCluster" and "NS.tendrl.flows.ImportCluster"  and not in "NS.node_agent",  this PR should go in commons.

What we do need to refactor is , use the NS.integrations.{ceph, gluster} namespace (https://github.com/Tendrl/node-agent/tree/develop/tendrl/integrations) for defining ceph, gluster specific atoms instead of current ceph_help, gluster_help like files in commons

shtripat:
>> these flows have to be in the root namespace which is "NS.tendrl" along with
>> NS.tendrl.flows.CreateCluster" and "NS.tendrl.flows.ImportCluster" and not in "NS.node_agent", this >> PR should go in commons.

>> What we do need to refactor is , use the NS.integrations.{ceph, gluster} namespace (https://github.com
>> /Tendrl/node-agent/tree/develop/tendrl/integrations) for defining ceph, gluster specific atoms instead >> of current ceph_help, gluster_help like files in commons

My understanding of flows is that any flow which is to be executed before existence of cluster, should be under `NS.tendrl` namespace and rest under either node-agent or integrations namespace.

The expand cluster is something which gets executed only once cluster is existing so ideal place is `NS.ceph` but due to restriction of ceph-installer here its been put under node-agent namespace.

Still I am open if you want it to be moved to commons.
Comments?

r0h4n:
The generic flows need to be in NS.tendrl, and when we start sds specific actions those should be moved to NS.integrations.ceph,  my parent comment contains more details about how we refactor these generic flows

shtripat:
Also the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1446396 is closed but with latest ceph-installer as well the mon addition fails. We are discussing with ceph team to resolve the same.

PR #419, opened 2017-05-10
Added a sample for invoking journal mapping flow

shtripat:

### DIFF
--- a/etc/generate_journal_mapping.py
+++ b/etc/generate_journal_mapping.py
@@ -1,0 +1,32 @@
+import json
+import uuid
+
+import etcd
+
+job_id1 = str(uuid.uuid4())
+
+job = {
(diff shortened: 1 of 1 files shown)


etc/generate_journal_mapping.py
```diff
@@ -0,0 +1,32 @@
...
+
+job = {
+    "run": "node_agent.flows.GenerateJournalMapping",
+    "status": "new",
+    "parameters": {
+        "Cluster.node_configuration": {
+            "12367fb3-0074-4b3d-aa90-b9518977987d": {
+              "storage_disks": [
+                {"device": "/dev/vdb", "size": 10737418240, "ssd": False},
+                {"device": "/dev/vdc", "size": 10737418240, "ssd": False},
```
r0h4n: Btw, we can detect if the disk is an ssd, do we need this flag?

PR #427, opened 2017-05-12
Added an object to store journal details while create cluster

shtripat:
This object is used from create cluster flow of commons module

### DIFF
--- a/tendrl/integrations/ceph/objects/journal/__init__.py
+++ b/tendrl/integrations/ceph/objects/journal/__init__.py
@@ -1,0 +1,20 @@
+from tendrl.commons import objects
+
+
+class Journal(objects.BaseObject):
+    internal = True
+    def __init__(self, integration_id, node_id, data, *args, **kwargs):
+        self._defs = {}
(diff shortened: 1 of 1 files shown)


tendrl/integrations/ceph/objects/journal/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Journal(objects.BaseObject):
+    internal = True
+    def __init__(self, integration_id, node_id, data, *args, **kwargs):
+        self._defs = {}
+        super(Journal, self).__init__(*args, **kwargs)
+
+        self.integration_id = integration_id
+        self.node_id = node_id
+        self.data = data
```
r0h4n: arent the journal details per OSD? I dont see any ref to OSD uuid here

shtripat: Per OSD journal details are journal metadata. This information I am maintaining per host for next time journal mapping process, so not more journals get created on SSD disks.

This is nothing to do with journal details of OSDs in ceph cluster.

PR #428, opened 2017-05-12
Consider journals already available on SSDs disk while mapping

shtripat:
Intended to invoke the journal mapping calculation wile expand
cluster flow as well. This would consider the already available
journals on any SSD with the node. If maximum no of possible
journals exceed the allowed limit, it would map accordingly.

While invocation from expand cluster flow the caller should pass
TendrlContext.integration_id as well so that existing details can
be considered.

### DIFF
--- a/tendrl/node_agent/flows/generate_journal_mapping/__init__.py
+++ b/tendrl/node_agent/flows/generate_journal_mapping/__init__.py
@@ -9,7 +9,8 @@ class GenerateJournalMapping(flows.BaseFlow):
     def run(self):
         # Generate the journal mapping for the nodes
         mapping = utils.generate_journal_mapping(
-            self.parameters['Cluster.node_configuration']
+            self.parameters['Cluster.node_configuration'],
+            integration_id=self.parameters.get("TendrlContext.integration_id")
         )
 
         # Update output dict
--- a/tendrl/node_agent/provisioner/ceph/utils.py
+++ b/tendrl/node_agent/provisioner/ceph/utils.py
@@ -1,3 +1,4 @@
+import json
 import os
 import struct
 import time
@@ -27,7 +28,7 @@ def generate_auth_key():
     return base64.b64encode(header + key).decode('utf-8')
 
 
-def generate_journal_mapping(node_configuration):
+def generate_journal_mapping(node_configuration, integration_id=None):
 
     """This fuction works based on a simple algorithm which maps
     smaller sized disks as journal for bigger sized disks.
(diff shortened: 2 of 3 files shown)


tendrl/node_agent/flows/generate_journal_mapping/__init__.py
```diff
@@ -9,7 +9,8 @@ class GenerateJournalMapping(flows.BaseFlow):
     def run(self):
         # Generate the journal mapping for the nodes
         mapping = utils.generate_journal_mapping(
-            self.parameters['Cluster.node_configuration']
+            self.parameters['Cluster.node_configuration'],
+            integration_id=self.parameters.get("TendrlContext.integration_id")
```
shtripat: Its an optional parameter. It should be passed if cluster is already existing and for additional disks we want a journal mapping. If not passed normal mapping would be done as usual. In case of existing cluster, while mapping it would try to check if SSDs allow more journals to be created on them.

r0h4n: generate_journal_mapping does not take more than 1 argument, youll need to fix that

shtripat: In same PR tendrl/node_agent/provisioner/ceph/utils.py#L31 has added an optional parameter for integration_id. Not sure what else is pending.

tendrl/node_agent/provisioner/ceph/utils.py
```diff
@@ -94,6 +95,17 @@ def generate_journal_mapping(node_configuration):
...
     for node_id in node_configuration.keys():
+
+        # Get the node specific journal details from central store
+        # if integration_id is passed. This is the case when this
+        # utility is called during expand cluster. We should check
+        # if pre-existing journal details available for the node
+        # and consider any SSD disk available for journal mapping
+        if integration_id:
+            journal_details = json.loads(NS.etcd_orm.client.read(
+                "clusters/%s/JournalDetails/%s/data" % (integration_id, node_id)
```
r0h4n: Create the Journal object and load it instead of raw read

shtripat: Will try that

r0h4n: Via commons, You can check if a Object(id=id).exists()  (returns True or false)

shtripat:
@r0h4n Verified the mapping part and it works as expected.
But ceph-installer has an issue where, if multiple OSD mapped to a single SSD for journal, expansion (adding OSD having journal on existing SSD disk) fails. I will be creating a BZ for that.

But this PR as is can taken in as it does only the mapping part.

r0h4n:
These flows should be moved to integrations.ceph Namespace, merging them for now @shtripat

PR #438, opened 2017-05-17
Added samples for expand cluster and add osd for ceph clusters

shtripat:

### DIFF
--- a/etc/add_osd_ceph_sample.py
+++ b/etc/add_osd_ceph_sample.py
@@ -1,0 +1,42 @@
+payload = {
+"run": "integrations.ceph.flows.AddOsds",
+"type": "node",
+"created_at": "2017-03-09T14:15:14Z",
+"username": "admin",
+"parameters": {
+    "TendrlContext.cluster_name": "MyCluster",
+    "TendrlContext.cluster_id": "140cd3d5-58e4-4935-a954-d946ceff371d",
+    "Cluster.public_network": "10.70.40.0/22",
+    "Cluster.cluster_network": "10.70.40.0/22",
+    "Cluster.node_configuration": {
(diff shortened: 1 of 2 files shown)


shtripat:

etc/add_osd_ceph_sample.py
```diff
@@ -0,0 +1,44 @@
...
+            "storage_disks": [{
+                "device": "/dev/vdf",
+                "journal": "/dev/vdg"
+            }
+            ]
+        },
+    },
+    "TendrlContext.integration_id": "89604c6b-2eff-4221-96b4-e41319240240"
+},
+"node_ids": [],
```
r0h4n: node_ids not required

etc/add_osd_ceph_sample.py
```diff
@@ -0,0 +1,44 @@
+payload = {
+"run": "integrations.ceph.flows.AddOsds",
+"type": "node",
+"created_from": "API",
```
r0h4n: remove created_from, only required for jobs coming from the API

shtripat: done

etc/add_osd_ceph_sample.py
shtripat: will remove that

etc/add_osd_ceph_sample.py
```diff
@@ -0,0 +1,42 @@
+payload = {
+"run": "integrations.ceph.flows.AddOsds",
+"type": "node",
+"created_at": "2017-03-09T14:15:14Z",
+"username": "admin",
+"parameters": {
+    "TendrlContext.cluster_name": "MyCluster",
```
r0h4n: TendrlContext.integration_id should be required

PR #440, opened 2017-05-17
Corrected an issue with journal object definition

shtripat:


shtripat:
@r0h4n plz review. Its a fix as per latest object definitions.

PR #447, opened 2017-05-18
Added logic to sync osd journal details while node sync

shtripat:
This is required if a cluster created from CLI is imported in
tendrl. We should persist the jorunal mapping details in central
store so that while expand cluster these details can be referenced

tendrl-bug-id: Tendrl/ceph-integration#236

### DIFF
--- a/tendrl/integrations/ceph/sds_sync/__init__.py
+++ b/tendrl/integrations/ceph/sds_sync/__init__.py
@@ -1,0 +1,82 @@
+import gevent
+import json
+import os
+
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message, ExceptionMessage
+from tendrl.commons.utils import cmd_utils
+from tendrl.commons import sds_sync
+from tendrl.commons.utils import log_utils as logger
+from tendrl.node_agent.node_sync import disk_sync
+
+
+class CephIntegrtaionsSyncThread(sds_sync.StateSyncThread):
+    def _run(self):
+        logger.log(
+            "debug",
+            NS.get("publisher_id", None),
+            {"message": "%s running" % self.__class__.__name__}
+        )
+        if NS.tendrl_context.integration_id is not None:
+            while not self._complete.is_set():
+                # sync of OSD journal details every 30 sec is fine
+                # as this data doesnt change very frequently in cluster
+                gevent.sleep(30)
+                try:
+                    # Sync the OSD journal details
+                    logger.log(
+                        "info",
+                        NS.get("publisher_id", None),
+                        {"message": "ceph_integrations_sync, osd journal details"}
+                    )
+                    journal_details = {}
+                    osd_path = "/var/lib/ceph/osd"
+                    osd_list = os.listdir("/var/lib/ceph/osd")
+
+                    if len(osd_list) > 0:
+                        for item in os.listdir(osd_path):
+                            journal_path = "%s/%s/journal" % (osd_path, item)
+                            out, err, rc = cmd_utils.Command(
+                                "lsblk -o PKNAME,ROTA -n %s" % journal_path
+                            ).run()
+                            if rc != 0:
+                                logger.log(
+                                    "error",
+                                    NS.get("publisher_id", None),
+                                    {"message": "Error getting journal details for OSD. Error: %s" % err}
+                                )
+                            else:
+                                journal_disk, rotational = out.split()
+                                if ("/dev/%s" % journal_disk) in journal_details.keys():
+                                    journal_details[
+                                        "/dev/%s" % journal_disk
+                                    ]['journal_count'] += 1
+                                    journal_details[
+                                        "/dev/%s" % journal_disk
+                                    ]['ssd'] = disk_sync.is_ssd(rotational)
+                                else:
+                                    journal_details["/dev/%s" % journal_disk] = {
+                                        'journal_count': 1,
+                                        'ssd': disk_sync.is_ssd(rotational)
+                                    }
+
+                        NS.integrations.ceph.objects.Journal(
+                            integration_id=NS.tendrl_context.integration_id,
+                            node_id=NS.node_context.node_id,
+                            data=json.dumps(journal_details)
+                        ).save(update=False)
+                except Exception as ex:
+                    logger.log(
+                        "error",
+                        NS.get("publisher_id", None),
+                        {
+                            "message": "ceph integrations sync failed: " + ex.message,
+                            "exception": ex
+                        }
+                    )
+
+            logger.log(
+                "debug",
+                NS.get("publisher_id", None),
+                {"message": "%s complete" % self.__class__.__name__}
+            )
--- a/tendrl/integrations/ceph/sds_sync/__init__.py
+++ b/tendrl/integrations/ceph/sds_sync/__init__.py
@@ -25,7 +25,7 @@ class CephIntegrtaionsSyncThread(sds_sync.StateSyncThread):
                 try:
                     # Sync the OSD journal details
                     logger.log(
-                        "info",
+                        "debug",
                         NS.get("publisher_id", None),
                         {"message": "ceph_integrations_sync, osd journal details"}
                     )
--- a/tendrl/node_agent/manager/__init__.py
+++ b/tendrl/node_agent/manager/__init__.py
@@ -19,6 +19,9 @@ from tendrl.node_agent.message.handler import MessageHandler
 from tendrl.node_agent import node_sync
 from tendrl import provisioning
 
+from tendrl.integrations.ceph import sds_sync as \
+    ceph_integrations_sds_sync
+
 
 class NodeAgentManager(commons_manager.Manager):
     def __init__(self):
@@ -95,6 +98,12 @@ def main():
     m = NodeAgentManager()
     m.start()
 
+    if NS.tendrl_context.sds_name == "ceph" and \
+        NS.tendrl_context.integration_id is not None:
+        NS.ceph_integrations_sync_thread = \
+            ceph_integrations_sds_sync.CephIntegrtaionsSyncThread()
+        NS.ceph_integrations_sync_thread.start()
+
     complete = gevent.event.Event()
 
     def shutdown():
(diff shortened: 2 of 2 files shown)


tendrl/node_agent/node_sync/__init__.py
```diff
@@ -358,6 +359,16 @@ def _run(self):
...
 
+                # Sync osd journal details
+                Event(
+                    Message(
+                        priority=priority,
+                        publisher=NS.publisher_id,
+                        payload={"message": "node_sync, osd journal details"}
+                    )
+                )
+                osd_sync.sync_osd_journal_details()
```
r0h4n: this should ideally go in integrations.ceph and it should have a sync class of its own. node_sync is generic and not related to an sds

shtripat: Done. Will move and verify the changes once.

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+                        priority="info",
+                        publisher=NS.publisher_id,
+                        payload={
+                            "message":
+                            "ceph_integrations_sync, osd journal details"
+                        }
+                    )
+                )
+                journal_details = {}
+                osd_list = os.listdir("/var/lib/ceph/osd")
```
r0h4n: can the user change this path ` /var/lib/ceph/osd` ? If yes, we need to handle that too.

Also why arent you simply using the osd-maps to sync the osd data in tendrl-ceph-integration itself instead of here?

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+                                    "/dev/%s" % journal_disk
+                                ]['ssd'] = disk_sync.is_ssd(rotational)
+                            else:
+                                journal_details["/dev/%s" % journal_disk] = {
+                                    'journal_count': 1,
+                                    'ssd': disk_sync.is_ssd(rotational)
+                                }
+
+                    NS.integrations.ceph.objects.Journal(
+                        integration_id=NS.tendrl_context.integration_id,
```
r0h4n: how are you dealing with the case where there's no NS.tendrl_context.integration_id yet? i.e the node is not registered to any Tendrl managed cluster

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: >> can the user change this path /var/lib/ceph/osd ? If yes, we need to handle that too.

I am not sure if this property can be changed.

>> Also why arent you simply using the osd-maps to sync the osd data in tendrl-ceph-integration
>> itself instead of here?

Here I want to update the OSDs for this OSD node only and file links available at `/var/lib/ceph/osd` provide the exact list.

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: May be we need to add a wait here for availability of `integration_id` under `tendrlcontext` for node?

tendrl/integrations/ceph/sds_sync/__init__.py
r0h4n: The property can be changed and the contents of /var/lib/ceph/osd can also have such dir structure /var/lib/ceph/osd/{cluster-name}/

http://docs.ceph.com/docs/master/install/manual-deployment/#adding-osds

tendrl/integrations/ceph/sds_sync/__init__.py
r0h4n: Thats a bad idea

This entire sync while loop should run only if NS.tendrl_context.integration_id is not None, this is the "integrations.$sds" namespace because it will require an integration id always

shtripat: Ok. will add that check.

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: Problem is while this sync we have node ids but in maps data we have osds listed as 0,1,2 etc. Not only that, the data we are looking for (destination disk for the journal of an OSD) is not available under maps data of the cluster.

r0h4n: - You can sync in integrations.ceph
- You dont need regex, just check for cluster name (default is "ceph")

shtripat: >> You can sync in integrations.ceph

I didnt get this point. Do you mean not say this as `tendrl/integrations/ceph/sds_sync` and rather say `tendrl/integrations/ceph/` ?

r0h4n: - Keep the current code location as is

- Dont use regex for /var/lib/osd parsing, just use "ceph" as default name or use whatever name is given to cluster in TendrlContext.cluster_name

shtripat: Ok. will update the PR with changes.

shtripat:
Verified

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+from tendrl.commons.utils import cmd_utils
+from tendrl.commons import sds_sync
+from tendrl.node_agent.node_sync import disk_sync
+
+
+class CephIntegrtaionsSyncThread(sds_sync.StateSyncThread):
+    def _run(self):
+        Event(
+            Message(
+                priority="info",
```
r0h4n: this should be debug, this isnt useful info for the user, this is more for the devs/support

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+        )
+        while not self._complete.is_set():
+            # sync of OSD journal details every 30 sec is fine
+            # as this data doesnt change very frequently in cluster
+            gevent.sleep(30)
+            if NS.tendrl_context.integration_id is not None:
+                try:
+                    # Sync the OSD journal details
+                    Event(
```
r0h4n: Please use https://github.com/Tendrl/commons/blob/develop/tendrl/commons/utils/log_utils.py instead of Event/Message

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+                            publisher=NS.publisher_id,
+                            payload={
+                                "message":
+                                "ceph_integrations_sync, osd journal details"
+                            }
+                        )
+                    )
+                    journal_details = {}
+                    if os.path.isdir(
+                        "/var/lib/ceph/osd/%s" % NS.tendrl_context.cluster_name
```
r0h4n: Can you please handle the case I had mentioned in previous reviews here

By default, Ceph expects that you will store a Ceph OSD Daemon’s data with the following path:

"/var/lib/ceph/osd/$cluster-$id" , where $cluster is TendrlContext.cluster_name and $id is the osd id, I dont see this PR handling it

http://docs.ceph.com/docs/jewel/rados/configuration/ceph-conf/#osds

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+                            payload={
+                                "message":
+                                "ceph integrations sync failed: " + ex.message,
+                                "exception": ex
+                            }
+                        )
+                    )
+            Event(
+                Message(
+                    priority="info",
```
r0h4n: debug

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: Done

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: Done

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: Done

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: Else part is taking of that only. It would list all the files under `/var/lib/ceph/osd` and the file names would be like `/var/lib/ceph/osd/{ClusterName}-{osd_id}`. Check lines 49,50 for the same.

tendrl/node_agent/manager/__init__.py
```diff
@@ -65,6 +68,8 @@ def main():
 
     NS.first_node_inventory_sync = True
     NS.state_sync_thread = node_sync.NodeAgentSyncThread()
+    NS.ceph_integrations_sync_thread = \
+        ceph_integrations_sds_sync.CephIntegrtaionsSyncThread()
```
shtripat: Done

tendrl/node_agent/manager/__init__.py
```diff
@@ -95,6 +98,10 @@ def main():
     m = NodeAgentManager()
     m.start()
 
+    NS.ceph_integrations_sync_thread = \
```
r0h4n: Please start this thread when

if NS.tendrl_context.sds_name == "ceph" and NS.tendrl_context.integration_id is not None:

shtripat: Ah yes. will change now.

tendrl/node_agent/manager/__init__.py
```diff
@@ -107,6 +116,7 @@ def shutdown():
         )
         complete.set()
         m.stop()
+        NS.ceph_integrations_sync_thread.stop()
```
r0h4n: wouldnt the same checks be required here  as start() ?

shtripat: Ah yes. sorry didnt realize this. adding now.

Issue #454, opened 2017-05-19
Network details are not came during node sync


GowthamShanmugam:
Disk id parsing logic have some problem, because of this problem node sync is end and came out, So network sync block is not executed, So network details are not pushed in etcd.

PR #455, opened 2017-05-19
Fix for getting disk id while parsing hwinfo output

GowthamShanmugam:
Disk id parsing logic have some error so disk sync gives
some problem. Becuase of this problem Network sysnc is not
happened.

tendrl-node-agent-bug-id: #454


GowthamShanmugam:
@r0h4n @shtripat @nthomas-redhat please review 

one more pull request in common is https://github.com/Tendrl/commons/pull/521

PR #468, opened 2017-05-22
Added logic to setup passwd less SSH to the OSD node

shtripat:
While add-osd to the existing OSD node, if the cluster is pure
import case (cluster created outside tendrl), we need to setup
passwd less SSH to the node so that ceph-installer can execute
the jobs for osd creation.

### DIFF
--- a/tendrl/integrations/ceph/flows/add_osds/__init__.py
+++ b/tendrl/integrations/ceph/flows/add_osds/__init__.py
@@ -1,3 +1,6 @@
+import gevent
+import uuid
+
 from tendrl.commons import flows
 from tendrl.commons.event import Event
 from tendrl.commons.flows.exceptions import FlowExecutionFailedError
@@ -5,6 +8,7 @@ from tendrl.commons.flows.create_cluster import \
     ceph_help as create_ceph_help
 from tendrl.commons.flows.expand_cluster import ceph_help
 from tendrl.commons.message import Message
+from tendrl.commons.objects.job import Job
 
 
 class AddOsds(flows.BaseFlow):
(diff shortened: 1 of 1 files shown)


tendrl/integrations/ceph/flows/add_osds/__init__.py
```diff
@@ -13,6 +17,64 @@ def run(self):
         if integration_id is None:
             raise FlowExecutionFailedError("TendrlContext.integration_id cannot be empty")
 
+        ssh_job_ids = []
+        ssh_setup_script = NS.ceph_provisioner.get_plugin().setup()
+        if len(self.parameters["Cluster.node_configuration"].keys()) > 0:
+            for node_id in self.parameters["Cluster.node_configuration"].keys():
+                new_params = {}
+                new_params['Node[]'] = [node_id]
```
r0h4n: what if current node is in Node[], you will have to setup ssh locally without a job

tendrl/integrations/ceph/flows/add_osds/__init__.py
```diff
@@ -13,6 +17,64 @@ def run(self):
         if integration_id is None:
             raise FlowExecutionFailedError("TendrlContext.integration_id cannot be empty")
 
+        ssh_job_ids = []
+        ssh_setup_script = NS.ceph_provisioner.get_plugin().setup()
+        if len(self.parameters["Cluster.node_configuration"].keys()) > 0:
+            for node_id in self.parameters["Cluster.node_configuration"].keys():
```
shtripat: Will remove the if check and `Cluster.node_configuration` is mandatory parameter for the flow.

tendrl/integrations/ceph/flows/add_osds/__init__.py
shtripat: In case of ceph, provisioner is outside the cluster so this scenario is not valid.

tendrl/integrations/ceph/flows/add_osds/__init__.py
shtripat: Anyway added a check for `Cluster.node_configuration` as well

r0h4n:
Verified?

shtripat:
Verified..

PR #470, opened 2017-05-23
Network out of band delete handle

GowthamShanmugam:

### DIFF
--- a/tendrl/node_agent/node_sync/__init__.py
+++ b/tendrl/node_agent/node_sync/__init__.py
@@ -226,26 +226,11 @@ class NodeAgentSyncThread(sds_sync.StateSyncThread):
                         payload={"message": "node_sync, Updating networks"}
                     )
                 )
-                # node wise network
-                try:
-                    NS._int.wclient.delete(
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@r0h4n @shtripat @nnDarshan please review

i have setting ttl for network instead of deleting, here what drawback i seen is, global networks are stored under 
/networks/{subnet_id}/{node_id}/{interface_id} so ttl is set for interface level, suppose node does not have interface for a particular subnet then /networks/{subnet_id}/{node_id}/ is Empty so i am deleting Empty node_id dir, same logic for  /networks/{subnet_id}/   , problem is if ttl comes 0 then extra interfaces are deleted but these two directory deletion will happen coming node sync only. so upto that /networks/{subnet_id}/{node_id}/Empty directory will present.

is this cause any problem

r0h4n:
Can you link to the tendrl issue?

GowthamShanmugam:
@r0h4n this is not an issue for this

GowthamShanmugam:
@r0h4n yes verified

PR #478, opened 2017-05-31
Added a patch for removing the glsuterfs_repo attr from conf file

nnDarshan:
Bug-URL: https://bugzilla.redhat.com/1456453


nnDarshan:

etc/patches/0001-remove-the-glusterfs_repo-var-from-conf.patch
```diff
@@ -0,0 +1,25 @@
+From 04a6f332205cb937f0ca89159176a3ff5c34fcee Mon Sep 17 00:00:00 2001
```
r0h4n: it should be "0001-use-default-platform-configured-gluster-repo.patch"

r0h4n:
Minor comment, plz fix

nnDarshan:
@r0h4n Done

etc/patches/0001-remove-the-glusterfs_repo-var-from-conf.patch
shtripat: In this .patch file actually we are removing the entry from conf file. Actually defaulting happens in python-gdeploy and create_cluster flow :)

Dont see anything wrong in name. BUT yes `whats in name ;)`

PR #507, opened 2017-06-15
Added logic to figure out the device path for OSD and add

shtripat:


shtripat:
Closing and would send from proper branch

PR #508, opened 2017-06-15
Added logic to figure out the device path for OSD and add

shtripat:
tendrl-bug-id: Tendrl/node-agent#510

### DIFF
--- a/tendrl/integrations/ceph/objects/definition/ceph.yaml
+++ b/tendrl/integrations/ceph/objects/definition/ceph.yaml
@@ -18,3 +18,76 @@ namespace.integrations.ceph:
       type: Create
       uuid: 2f94a48a-05d7-408c-b400-e27827f4edcd
       version: 1
+  objects:
+    Osd:
+      attrs:
+        id:
+          help: Id of the OSD
+          type: int
+        uuid:
+          help: UUID
+          type: String
+        hostname:
+          help: Name of the host
+          type: String
+        public_addr:
+          help: Public Address
+          type: String
+        cluster_addr:
+          help: Cluster Address
+          type: String
+        device_path:
+          help: device path
+          type: String
+        heartbeat_front_addr:
+          help: Heartbeat front address
+          type: String
+        heartbeat_back_addr:
+          help: Heartbeat back address
+          type: String
+        down_at:
+          help: Down at
+          type: int
+        up_from:
+          help: Up from
+          type: int
+        lost_at:
+          help: Lost at
+          type: int
+        osd_up:
+          help: OSD up flag
+          type: int
+        osd_in:
+          help: OSD int flag
+          type: int
+        up_thru:
+          help: Up through
+          type: int
+        weight:
+          help: Weight
+          type: String
+        primary_affinity:
+          help: Primary affinity
+          type: String
+        state:
+          help: State
+          type: String[]
+        last_clean_begin:
+          help: Last clean begin
+          type: int
+        last_clean_end:
+          help: Last clean end
+          type: int
+        total:
+          help: total size of osd
+          type: int
+        used:
+          help: used size
+          type: int
+        used_pcnt:
+          help: used percent
+          type: String
+      help: "Osd"
+      enabled: true
+      value: clusters/$TendrlContext.integration_id/Osds/$Osd.id
+      list: clusters/$TendrlContext.integration_id/Osds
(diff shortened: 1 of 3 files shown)


r0h4n:
Please an issue describing what this PR is for

shtripat:
Added the issue id

tendrl/integrations/ceph/objects/osd/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+        self.last_clean_end = last_clean_end
+        self.total=total
+        self.used=used
+        self.used_pcnt=used_pcnt
+        self.value = 'clusters/{0}/Osds/{1}'
+
+    def render(self):
+        self.value = self.value.format(
+            NS.tendrl_context.integration_id,
+            self.id
```
r0h4n: use uuid instead of id

tendrl/integrations/ceph/sds_sync/__init__.py
```diff
@@ -23,6 +23,26 @@ def _run(self):
...
+                    # Sync the OSD device path details
+
+                    # A sample output from df command would be as below
+                    # /dev/vdb1 10474476 3547584 6926892 34% /var/lib/ceph/osd/ceph_20175131497330322341-3
+                    # /dev/vdc1 10474476 1036704 9437772 10% /var/lib/ceph/osd/ceph_20175131497330322341-2
+
+                    pattern = "/var/lib/ceph/osd/%s-" % NS.tendrl_context.cluster_name
+                    out, err ,rc = cmd_utils.Command(('df')).run()
+                    if not err and rc == 0:
+                        for line in out.split('\n'):
```
r0h4n: check empty "out"

tendrl/integrations/ceph/objects/osd/__init__.py
shtripat: Earlier it was `uuid`. The only thing is while syncing the device path for the OSDs we dont have `uuid` available and to update the details we would need to loop through all the OSDs to set one field value. Because of this I changed it to identify the OSDs by their `id` and not `uuid`. Under a specific cluster anyways this no is going to be unique.

tendrl/integrations/ceph/sds_sync/__init__.py
shtripat: done

tendrl/integrations/ceph/objects/osd/__init__.py
r0h4n: Move definition of Osd from here https://github.com/Tendrl/ceph-integration/blob/develop/tendrl/ceph_integration/objects/definition/ceph.yaml#L597 

to this NS definition yaml and remove the uuid from "value" in defs

Dunno how you tested this successful without definitions updated

shtripat: My bad I forgot to add the ceph.yaml file modified under integration. The change is done and tested with that only.

Also dont we need definition of Osd under `ceph-integration/objects/definition/ceph.yaml`  after this change?

r0h4n: Yes we need to move object definition to ceph-integration/objects/definition/ceph.yaml

how did you test this without adding definition to ceph-integration/objects/definition/ceph.yaml  was my question

shtripat: I had added to integration/ceph/objects/definition/ceph.yaml and tested with that only. Forgot to add the file in PR somehow.

Also if we move the definition of Osd from ceph-integration to this, how would be able to add flows and atoms for Osd later in ceph-integration?

tendrl/integrations/ceph/objects/definition/ceph.yaml
```diff
@@ -18,3 +18,76 @@ namespace.integrations.ceph:
...
+          type: int
+        used:
+          help: used size
+          type: int
+        used_pcnt:
+          help: used percent
+          type: String
+      help: "Osd"
+      enabled: true
+      value: clusters/$TendrlContext.integration_id/Osds/$Osd.uuid
```
r0h4n: I had asked you to remove uuid, there's no rush, read comments and then implement

tendrl/integrations/ceph/objects/osd/__init__.py
r0h4n: You will have to duplicate the definition of OSD object in NS.ceph and NS.integrations.ceph

and then add flows,atoms to NS.ceph

tendrl/integrations/ceph/objects/definition/ceph.yaml
shtripat: Done

PR #517, opened 2017-06-19
Update Readme.rst

rishubhjain:
Experienced some issues while following the guidelines, the following changes solved those issues.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,7 +131,7 @@ Configuration
 
    ::
 
-       $ cp etc/tendrl/node-agent/node-agent-dev.conf.yaml /etc/tendrl/node-agent/node-agent.conf.yaml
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -143,7 +143,7 @@ Configuration
    ::
 
        $ cp node-agent/etc/tendrl/node-agent/logging.yaml.timedrotation.sample \
-         /etc/tendrl/node-agent_logging.yaml
```
anmolbabu: This is no longer required... 
I think you can remove it totally
cc: @r0h4n

r0h4n: - Delete the file `logging.yaml.timedrotation.sample`

- Use `logging.yaml.syslog.sample`

PR #527, opened 2017-06-23
Added logic to set tag provisioner/{integration-id} for ceph case

shtripat:


shtripat:
not verified yet

shtripat:
Not required and added mechanism to target the job using tags.
https://github.com/Tendrl/commons/pull/628

Issue #522, opened 2017-06-21
disk raw_reference is common for virtual and physical disk


GowthamShanmugam:
Now we are storing disk raw reference in /nodes/{node_id}/LocalStorage/Disks/RawReference, but virtual disks also have disk raw_reference, i feel it has to store in  /nodes/{node_id}/LocalStorage/RawReference. if the node is virtual machine then raw_reference contains virtual disk detail, if node is physical device then raw_reference should have physical device details.

GowthamShanmugam:
@brainfunked @shtripat @r0h4n

shtripat:
Looks reasonable to me.

r0h4n:
@GowthamShanmugam How are the current RawReference used by API and UI?

brainfunked:
I agree with the change. I don't think the file is being used by either the API or the UI presently. I had asked for it to be synced for future use.

GowthamShanmugam:
yes it is not used by anyone

GowthamShanmugam:
@r0h4n We can close this issue

PR #528, opened 2017-06-23
Disk raw_reference is moved out form etcd disk directory

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#522


GowthamShanmugam:
@r0h4n please review

PR #534, opened 2017-06-28
Modified to use tags instead of node_ids in flows

shtripat:
tendrl-bug-id: Tendrl/commons#603

### DIFF
--- a/tendrl/integrations/ceph/flows/add_osds/__init__.py
+++ b/tendrl/integrations/ceph/flows/add_osds/__init__.py
@@ -28,7 +28,7 @@ class AddOsds(flows.BaseFlow):
             new_params['Node[]'] = [node_id]
             new_params['ssh_setup_script'] = ssh_setup_script
             payload = {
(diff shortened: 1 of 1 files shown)


shtripat:
Need to be verified

tendrl/integrations/ceph/flows/add_osds/__init__.py
```diff
@@ -28,7 +28,7 @@ def run(self):
             new_params['Node[]'] = [node_id]
             new_params['ssh_setup_script'] = ssh_setup_script
             payload = {
-                "node_ids": [node_id],
+                "tags": ["tendrl/node_%s" % node],
```
r0h4n: it is "node_id" and not "node"

shtripat: ah yes. will change that

shtripat:
Verified the changes with add osd flow and works as expected.

Issue #538, opened 2017-06-30
Disk sync raise exception when disk id contain space


GowthamShanmugam:
Physical disk id is a combination of device+vendor+serail_no, Some times hwinfo gives space in device or vendor. so because of this problem etcd key has space and giving etcd error also. 
Need to remove spaces from device and vendor and serial_no to solve this problem.

GowthamShanmugam:
@dahorak @r0h4n

PR #539, opened 2017-06-30
Removing spaces form disk id

GowthamShanmugam:
tendrl-bug-id : Tendrl/node-agent#538

### DIFF
--- a/tendrl/node_agent/node_sync/disk_sync.py
+++ b/tendrl/node_agent/node_sync/disk_sync.py
@@ -170,16 +170,16 @@ def get_disk_details():
                         disk.split(':')[1].lstrip().replace('"', "")
                 elif key.strip() == "Vendor":
                     devlist["vendor"] = \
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@brainfunked but key combination already have underscore is that ok.
device+"_"+vendor+"_"+serial_id

brainfunked:
That'll need to be changed then. The character joining the words should be relatively obscure that it'll generally not be a part of a standard name. Colon maybe?

GowthamShanmugam:
@r0h4n is colon is usable for key ?

GowthamShanmugam:
@r0h4n need this patch for solve disk id problem which is faced by dahorak

r0h4n:
@brainfunked @GowthamShanmugam please revisit the disk object to improve the naming scheme https://github.com/Tendrl/commons/blob/develop/tendrl/commons/objects/disk/__init__.py#L42

GowthamShanmugam:
@r0h4n ok

PR #545, opened 2017-07-26
Changes for performance improvement

shtripat:
tendrl-spec: Tendrl/specifications#174


shtripat:
This PR is not required with latest changes

PR #555, opened 2017-08-03
Spec/refactoring of node monitoring flows into node agent

anmolbabu:


anmolbabu:
@nthomas-redhat @shtripat @r0h4n Please review..
Will keep adding fixes if any to this PR..

PR #565, opened 2017-08-04
Add up counters

anmolbabu:


anmolbabu:
@nthomas-redhat @shtripat @r0h4n Please review

tendrl/node_agent/monitoring/collectd/collectors/gluster/utils.py
```diff
@@ -1,7 +1,7 @@
 import ConfigParser
+from gevent import socket
```
shtripat: What this change is meant for ?

anmolbabu: socket calls are blocking calls. So I feel anyway our plugins are green threads so lets import socket also from gevent so that is compatible with gevent..

@r0h4n Also https://github.com/Tendrl/specifications/pull/206/files#diff-50d5b9e1a218ac822a65c48d23c48d57R138 is the spec that talks about making plugins gevent..Plus this idea is also in ceph-metrics project the only difference is they are using normal threads while we have more light-weight gevents...

r0h4n:
Please attach a detailed issue for this PR

PR #584, opened 2017-08-22
Fix heal info parsing

anmolbabu:
The issue was accessing a non-existant key fixed that
by storing brick index in get-state parsed result.
The other issue was in pushing values of bricks from
all nodes which caused the issue of node-name mismatch
when gluster hostname is different from tendrl identified
fqdn


anmolbabu:
@nthomas-redhat @shtripat @r0h4n Please review

anmolbabu:
Verified

PR #586, opened 2017-08-23
Modified node-agent to push the alert into alerting directory in etcd

GowthamShanmugam:
tendrl-spec-issue: Tendrl/specifications#169


GowthamShanmugam:
@anmolbabu @shtripat @r0h4n @rishubhjain @nthomas-redhat please review

GowthamShanmugam:
@shtripat @r0h4n @anmolbabu please review

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,122 @@
...
+                ack_comment=alert.ack_comment,
+                acked_at = alert.acked_at,
+                pid=alert.pid,
+                source=alert.source,
+                delivered=alert.delivered
+            ).save()
+
+def update_alert_count(alert, existing_alert):
+    counter_objects = []
+    if alert.tags['alert_catagory'] == "node":
```
anmolbabu: Who sets this alert_catagory ?
Will the answer vary depending on source of alert like:
* Grafana
* gluster-integration
So seeing this code I am of the feeling that whoever is pushing an alert to node-agent message socket should populate this. I know, now we make sure that alert is processed into a Tendrl alert before being pushed into node-agent socket. But can you link here to how its done in code is it automatic possibly by some classification of handlers into possibly 2 categories node and cluster and the base class auto populates this as soon as a handler plugin is added to a specific category? I am just thinking about existence of a safe default for this

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,122 @@
...
+        node_counter_obj = NodeAlertCounters(
+            node_id=alert.node_id).load()
+        counter_objects.append(node_counter_obj)
+    elif alert.tags['alert_catagory'] == "cluster":
+        cluster_counter_obj = ClusterAlertCounters(
+            integration_id=alert.tags['integration_id']
+        ).load()
+        counter_objects.append(cluster_counter_obj)
+        if alert.resource == "volume_utilization_alert":
```
anmolbabu: Hard coding gluster terminology in node-agent is probably not good.
Also you need to acquire lock from the stage of reading to the stage of updating as you can expect different node-agents to be simultaneously trying to do this update for different volumes(I understand duplicate alerts for same resource is not a possibility due to the design of alert cycle)

tendrl/node_agent/objects/volume_alert_counters/__init__.py
```diff
@@ -0,0 +1,24 @@
+from tendrl.commons import objects
+
+
+class VolumeAlertCounters(objects.BaseObject):
```
anmolbabu: I am somehow not able to convince myself with usage of sds-specific terminologies in a non sds module like node-agent unless absolutely necessary as in case of sds-detection etc..
Can this be achieved by some other means?

tendrl/node_agent/alert/utils.py
GowthamShanmugam: As per our discussion, i will change this alert_catagory as a mandatory parameter.

tendrl/node_agent/alert/utils.py
GowthamShanmugam: yes i feel the same but problem is i have to identify the alert is volume related alert or not so here i need something to identify.

and i have created volume counter object also in node-agent, i feel that is also not a good approach but i need volume counter object in node-agent to store volume alert count. am i move that object to common ?

tendrl/node_agent/objects/volume_alert_counters/__init__.py
GowthamShanmugam: am i move this to Commons because node-agent have responsible to find all alert counts so it should be accessible by node-agent

tendrl/node_agent/alert/utils.py
anmolbabu: @r0h4n @brainfunked What do you suggest?

tendrl/node_agent/objects/volume_alert_counters/__init__.py
anmolbabu: No don't move this to commons..

tendrl/node_agent/alert/utils.py
GowthamShanmugam: @r0h4n need your suggestion

tendrl/node_agent/objects/volume_alert_counters/__init__.py
GowthamShanmugam: @r0h4n need your suggestion

tendrl/node_agent/alert/utils.py
GowthamShanmugam: @anmolbabu yes it is like repetitive, one more problem is @brainfunked told like only node-agent has to read and write alerts in  etcd. if i follow the above approach then monitoring-integration also has to read alerts from etcd and check which is cleared event or new event or existing event with the same severity.

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,123 @@
...
+    if alert.classification == "node":
+        node_counter_obj = NodeAlertCounters(
+            node_id=alert.node_id).load()
+        counter_objects.append(node_counter_obj)
+    elif alert.classification == "cluster":
+        cluster_counter_obj = ClusterAlertCounters(
+            integration_id=alert.tags['integration_id']
+        ).load()
+        counter_objects.append(cluster_counter_obj)
+        if alert.resource == "volume_utilization_alert":
```
r0h4n: Gowtham, put all the sds specific code in https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/$sds/

and simply call tendrl.integrations.gluster.alerts.update_alert_count at L99

tendrl/node_agent/objects/definition/node_agent.yaml
```diff
@@ -67,4 +67,52 @@ namespace.node_agent:
...
+        info_count:
+          help: "Count of ok alerts for the cluster"
+          type: int
+        integration_id:
+          help: "Id of the cluster"
+          type: String
+      value: /clusters/{0}/alert_counters
+      list: /clusters/{0}/alert_counters
+      help: "Cluster Alert Counter"
+    VolumeAlertCounters:
```
r0h4n: Define and implement this object at https://github.com/Tendrl/node-agent/blob/master/tendrl/integrations/gluster/objects/definition/gluster.yaml

Remove from here

tendrl/node_agent/alert/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        new_alert_obj = AlertUtils().to_obj(new_alert)
+        if not new_alert_obj.alert_type in SUPPORTED_ALERT_TYPES:
+            logger.log(
+                "error",
+                NS.publisher_id,
+                {
+                    "message": "Invalid alert type in alert %s" % alert_str
+                }    
+            )
+            raise InvalidAlertType
```
shtripat: An instance of `InvalidAlertType` ?

tendrl/node_agent/alert/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        KeyError,
+        InvalidAlertType, 
+        EtcdKeyNotFound,
+        EtcdException
+    ) as ex:
+        logger.log(
+            "error",
+            NS.publisher_id,
+            {
+                "message": "Error in updating alert %s" % alert_str
```
shtripat: Do we know the why part of this error? If so we should add to the error message.

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,123 @@
...
+            delivered=alert.delivered
+        ).save()
+
+def update_alert_count(alert, existing_alert):
+    counter_objects = []
+    if alert.classification == "node":
+        node_counter_obj = NodeAlertCounters(
+            node_id=alert.node_id).load()
+        counter_objects.append(node_counter_obj)
+    elif alert.classification == "cluster":
```
shtripat: Create constants for alert classifications types and use them

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,123 @@
...
+                integration_id=alert.tags['integration_id'],
+                volume_id=alert.tags['volume_id']
+            ).load()
+            counter_objects.append(volume_counter_obj)
+
+    for counter_obj in counter_objects:
+        warn_count = int(counter_obj.warning_count)
+        info_count = int(counter_obj.info_count)
+        if existing_alert:
+            if alert.severity == "INFO":
```
shtripat: Create constants for `INFO`, `WARNING` etc and use

tendrl/node_agent/message/handler.py
```diff
@@ -35,6 +35,12 @@ def read_socket(self, sock, *args):
             frmt = "=%ds" % size
             msg = struct.unpack(frmt, data)
             message = Message.from_json(msg[0])
+            # Logger is in commons so passing alert from here
+            if message.priority == "notice":
```
shtripat: Constant for `notice` type

tendrl/node_agent/objects/volume_alert_counters/__init__.py
shtripat: As suggested by @r0h4n earlier move them under integrations and use

tendrl/node_agent/alert/__init__.py
GowthamShanmugam: it is class name, we can raise exception using class name

tendrl/node_agent/alert/__init__.py
GowthamShanmugam: in all the place before raising alert i am logging exact error message so using that we can identify easily

GowthamShanmugam:
@shtripat @anmolbabu @r0h4n all changes done

tendrl/integrations/gluster/objects/definition/gluster.yaml
```diff
@@ -1,26 +1,22 @@
 ---
 namespace.integrations.gluster:
-  flows:
```
anmolbabu: Accidentally removed?

tendrl/integrations/gluster/utils.py
```diff
@@ -0,0 +1,22 @@
+def update_volume_alert_count(alert, existing_alert):
+    counter_obj = NS.integrations.gluster.objects.VolumeAlertCounters(
+        integration_id=alert.tags['integration_id'],
+        volume_id=alert.tags['volume_id']
+    ).load()
+    warn_count = int(counter_obj.warning_count)
+    info_count = int(counter_obj.info_count)
+    if existing_alert:
+        if alert.severity == "INFO":
```
anmolbabu: Do you really need this?
I feel just increment or decrement  the alert count.
Also if you remove info, you can call it just alert count.

tendrl/integrations/gluster/utils.py
```diff
@@ -0,0 +1,22 @@
...
+            warn_count -= 1
+            info_count += 1
```
anmolbabu: Do you really need this?
I feel just increment or decrement the alert count.
Also if you remove info, you can call it just alert count.

tendrl/integrations/gluster/utils.py
```diff
@@ -0,0 +1,22 @@
+def update_volume_alert_count(alert, existing_alert):
```
anmolbabu: You probably have https://github.com/Tendrl/node-agent/pull/586/files#diff-fc525d4bb65f3b413979c834c12b06d0R95 so is this required?

tendrl/node_agent/objects/definition/node_agent.yaml
```diff
@@ -67,4 +67,34 @@ namespace.node_agent:
       list: nodes/$NodeContext.node.id/Networks
       help: "Global subnet wise network interface"
       value: /networks
+    ClusterAlertCounters:
+      enabled: True
+      attrs:
+        warning_count:
+          help: "Count of warning alerts for the cluster"
+          type: int
+        info_count:
```
anmolbabu: You probably don't need this.
Just AlertCounter suffices
Same in all counters

tendrl/integrations/gluster/objects/definition/gluster.yaml
GowthamShanmugam: @shtripat told this is just sample and this cause issue when i run code so he suggested me to remove it

anmolbabu: Ok I thought it was accidentally removed

tendrl/integrations/gluster/utils.py
GowthamShanmugam: This function is for calculate volume alert count, so i implemented here , this https://github.com/Tendrl/node-agent/pull/586/files#diff-fc525d4bb65f3b413979c834c12b06d0R95 will call
update_volume_alert_count() to calculate volume alert count

tendrl/node_agent/objects/definition/node_agent.yaml
GowthamShanmugam: @shtripat is info count for alert is required ?

tendrl/integrations/gluster/utils.py
GowthamShanmugam: @shtripat is info count for alert is required ?

tendrl/integrations/gluster/utils.py
anmolbabu: But why is this required?
I see everythin here except counter_obj creation is there in https://github.com/Tendrl/node-agent/pull/586/files#diff-fc525d4bb65f3b413979c834c12b06d0R95 can you move that counter_obj creation there?

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,122 @@
...
+       counter_obj = NodeAlertCounters(
+            node_id=alert.node_id).load()
+    elif alert.classification == CLUSTER_ALERT:
+        counter_obj = ClusterAlertCounters(
+            integration_id=alert.tags['integration_id']
+        ).load()
+        if alert.resource == "volume_utilization_alert":
+            utils.update_volume_alert_count(alert, existing_alert)
+
+    warn_count = int(counter_obj.warning_count)
```
anmolbabu: How these would behave in case of volume_utilization_alert ?
Does it not catch AttributeError?

GowthamShanmugam: modified please check

tendrl/node_agent/objects/definition/node_agent.yaml
GowthamShanmugam: i removed it

tendrl/integrations/gluster/utils.py
GowthamShanmugam: modified this logic

GowthamShanmugam:
@anmolbabu @r0h4n all changes done

GowthamShanmugam:
@nthomas-redhat i had discussed with @brainfunked about that document, he told me like keep the current way of implementation. Because the implementation which is proposed in that document needs some clarification.

GowthamShanmugam:
@r0h4n changes done please review

GowthamShanmugam:
alert implementation steps : https://github.com/Tendrl/specifications/pull/217#issuecomment-326881952

tendrl/node_agent/message/handler.py
```diff
@@ -35,6 +36,12 @@ def read_socket(self, sock, *args):
             frmt = "=%ds" % size
             msg = struct.unpack(frmt, data)
             message = Message.from_json(msg[0])
+            # Logger is in commons so passing alert from here
+            if message.priority == ALERT_PRIORITY:
+                update_alert(
```
r0h4n: define "update_alert" for tendrl.integrations.gluster.alert and call that function too from here.

eg:  node_agent.alert.update_alert
       integrations.gluster.alert.update_alert

tendrl/node_agent/alert/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    if alert.classification == constants.NODE_ALERT:
+        counter_obj = NodeAlertCounters(
+            node_id=alert.node_id).load()
+        alert_count_obj.append(counter_obj) 
+    elif alert.classification == constants.CLUSTER_ALERT:
+        counter_obj = ClusterAlertCounters(
+            integration_id=alert.tags['integration_id']
+        ).load()
+        alert_count_obj.append(counter_obj)
+        if alert.resource == constants.VOLUME_ALERT:
```
r0h4n: remove the VOLUME_ALERT check and move it o integrations.gluster.alerts.update_alert_count

etc/tendrl/node-agent/node-agent.conf.yaml.sample
```diff
@@ -6,6 +6,7 @@ log_level: DEBUG
 package_source_type: rpm
 logging_socket_path: /var/run/tendrl/message.sock
 message_retention_time: 172800 # default 2 days(2*24*60*60)
+clearing_alert_time: 172800 # default 2 days(2*24*60*60)
```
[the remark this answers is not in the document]
GowthamShanmugam: i will change this

tendrl/node_agent/message/handler.py
GowthamShanmugam: @r0h4n why we need one more update_alert function in gluster-integration. Except for volume alert count calculation, there is no integration related operation are happening in node-agent update_alert function.


update_alert function just read /alerting/nodes or /alert/clusters directory in etcd and compare the alert with new incoming alert and update  in etcd. 

other than this it will find alert count of host level alert count and cluster level alert count. and volume level alert count. 
Here only volume alert count is integration specific. other than this nothing is integration specific.

GowthamShanmugam: for this change, same logic repeats twice

GowthamShanmugam:
@r0h4n changes done

GowthamShanmugam:
@r0h4n thanks, i will fix pep8

tox.ini
```diff
@@ -72,7 +72,7 @@ commands =
 # E123, E125 skipped as they are invalid PEP-8.
 # see: http://flake8.readthedocs.io/en/latest/config.html#settings
 show-source = True
-ignore = E123,E125
+ignore = E123,E125,E402
```
GowthamShanmugam: skipping this error because 
it is import not in top error 

anmol used logic here https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_utilization.py#L12

import between sys so it isvalid case so i am skipping this error

PR #587, opened 2017-08-23
Configure profile info and heal statistics only on provisioner node

anmolbabu:
Configure profile info and heal statistics only on provisioner node and
hence, the same plugin will push stats of all bricks of all nodes
participating in volume.


anmolbabu:
@shtripat @nthomas-redhat @r0h4n Please review.

Needs to be verified

r0h4n:
Youll need to change the `tendrl.objects.Cluster.atoms.ConfigureMonitoring` atom to selectively configure these plugins only on "provisioner/$integration_id" node

anmolbabu:
@shtripat YEs that's there on commons as in https://github.com/Tendrl/commons/pull/697

Issue #590, opened 2017-08-24
Avoid another transaction in progress gluster error


anmolbabu:
This can be achieved by:
1. Varying the interval of plugin execution to a prime number such that
   the gluster-integration invocation of cluster level commands does not
   (or very least frequently) interleaves with collectd's gluster plugins
   invoking gluster cluster level commands.
2. Merging profile info and heal statistics plugins so that they are not
   executed in parallel by any chance among collectd plugins.
3. Add 3 retrial mechanisms and fail only after 3 attempts so that even if
   once the command fails, the probability of plugin failure is reduced
   drastically.

anmolbabu:
Closing this as PRs to fix this have been merged..
Will re-open if the issue re-occurs

PR #591, opened 2017-08-24
Avoid another transaction in progress gluster error

anmolbabu:
This pr achieves this by:
1. Varying the interval of plugin execution to a prime number such that
   the gluster-integration invocation of cluster level commands does not
   (or very least frequently) interleaves with collectd's gluster plugins
   invoking gluster cluster level commands.
2. Merging profile info and heal statistics plugins so that they are not
   executed in parallel by any chance among collectd plugins.
3. Add 3 retrial mechanisms and fail only after 3 attempts so that even if
   once the command fails, the probability of plugin failure is reduced
   drastically.

tendrl-bug-id: Tendrl/node-agent#590


anmolbabu:
@nthomas-redhat @shtripat @r0h4n Please review.

This seems to be working But we get to know actual results only after waiting for some more time

PR #593, opened 2017-08-30
Fixed loading of node context before update

shtripat:
tendrl-bug-id: Tendrl/ui#528


shtripat:
@r0h4n plz review

Issue #595, opened 2017-08-31
Set the tag `tendrl/node_{node-id}` while sds sync sart only


shtripat:
The node-id is something which doesn't change during the course of tendrl-node-agent sync cycles. Populate the tag `tendrl/node_{node-id}` while starting of node_sync of `tendrl-node-agent`.

shtripat:
Closing this as required changes are merged

PR #596, opened 2017-08-31
Moved the adding of tag tendrl/node_{node-id} in start of node sync

shtripat:
tendrl-bug-id: Tendrl/node-agent#595


shtripat:
@r0h4n plz review

PR #601, opened 2017-09-05
[WIP] Primary Selinux policy for tendrl

TimothyAsirJeyasing:
This patch provides initial selinux policy for tendrl.
The initial tendrl selinux policies are running as permissive domains
Which does the following things:
 - Allows tendrl process to run in permissive mode
 - Checks are performed but not enforced
 - Users don't have to switch to permissive mode globally
 - Allows to catch AVC messages
 - Once all the AVC are fixed, we can move to enfore later on.


TimothyAsirJeyasing:
@r0h4n , @nthomas-redhat , @shtripat , Please review

Issue #606, opened 2017-09-07
Add iops, throughput, latency and utilization of disks in the bricks


anmolbabu:
Add collectd plugin to fetch iops, throughput, latency and utilization of disks in the bricks

anmolbabu:
The PR to add these is merged.
So closing this issue

PR #607, opened 2017-09-07
Add brick to disk linking

anmolbabu:
tendrl-bug-id: Tendrl/node-agent#606

### DIFF
--- a/tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
+++ b/tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
@@ -1,0 +1,421 @@
+import psutil
+import socket
+import threading
+import time
+import traceback
+
+
+import collectd
+from tendrl_gluster import TendrlGlusterfsMonitoringBase
+import utils as gluster_utils
+
+
+class TendrlBrickDeviceStatsPlugin(
+    TendrlGlusterfsMonitoringBase
+):
+    def __init__(self):
+        self.provisioner_only_plugin = False
+        TendrlGlusterfsMonitoringBase.__init__(self)
+        self.STAT_INTERVAL_FOR_PER_SEC_COUNTER = 10
+        self.brick_details = {}
+
+    def get_brick_source_and_mount(self, brick_path):
+        # source and target correspond to fields "Filesystem" and
+        # "Mounted on" from df command output. The below command
+        # gives the filesystem and mount point for a given path,
+        # Eg: "/dev/mapper/tendrlMyBrick4_vg-tendrlMyBrick4_lv " \
+        #     "/tendrl_gluster_bricks/MyBrick4_mount"
+
+        command = "df --output=source,target " + brick_path
+        out, err = gluster_utils.exec_command(command)
+        if err:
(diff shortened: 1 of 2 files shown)


anmolbabu:
@r0h4n @shtripat @nthomas-redhat Please review

anmolbabu:
Verified

anmolbabu:
@r0h4n Done. Updated PR with pep8 fixes

tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
```diff
@@ -0,0 +1,421 @@
...
+                self.brick_details[
+                    'clusters.%s.volumes.%s.nodes.%s.bricks.%s.device.%s.'
+                    'disk_octets.read' % (
+                        self.CONFIG['integration_id'],
+                        vol_name,
+                        brick_host.replace('.', '_'),
+                        brick_path.replace('/', '|'),
+                        brick_device.replace('/dev/', '')
+                    )
+                ] = self.get_interval_disk_io_stat(
```
nthomas-redhat: @anmolbabu , why you are making these method calls again and again(same as in line 131). Instead you can make the call once use it at both places right? Is there specific reason why you are doing like this?

I think the same is applicable at multiple places in this method

tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
```diff
@@ -0,0 +1,421 @@
...
+
+    def get_brick_source_and_mount(self, brick_path):
+        # source and target correspond to fields "Filesystem" and
+        # "Mounted on" from df command output. The below command
+        # gives the filesystem and mount point for a given path,
+        # Eg: "/dev/mapper/tendrlMyBrick4_vg-tendrlMyBrick4_lv " \
+        #     "/tendrl_gluster_bricks/MyBrick4_mount"
+
+        command = "df --output=source,target " + brick_path
+        out, err = gluster_utils.exec_command(command)
```
shtripat: Cant we use `cmd_utils` from commons here?

tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
anmolbabu: There's not much gain right..
I mean the info even now is fetched only once but diff calculations are happening on demand

tendrl/node_agent/monitoring/collectd/collectors/gluster/low_weight/tendrl_glusterfs_brick_disk_stats.py
anmolbabu: I thought about it but that would make the plugins tied to tendrl but now except for config the plugins are independent and don't have any dependency on any tendrl module

shtripat: Ack

Issue #615, opened 2017-09-12
Nodecontext tags are not updated


nthomas-redhat:
Service sync has issues, tags are updated as expected

shtripat:
@nthomas-redhat this can be closed now.

PR #616, opened 2017-09-12
Corrected the sorting logic of tags

shtripat:
The function sort() return any value and rather updates the list
content itself. So we need to execute sort() explicityl on the
lists and then compare.

tendrl-bug-id: Tendrl/node-agent#615


shtripat:
@r0h4n @nthomas-redhat the regression issue while syncing the service tags for nodes.

Issue #619, opened 2017-09-13
Collectd does not report disk stats for brick created from partitioned disk


anmolbabu:
The reason for this are as  follows:
1. gluster-integration syncs only name of the block device corresponding to the brick and does not sync details of the partition of the disk from which the brick is carved out.
2. Collectd plugin for gathering disk stats for a brick relies on devices maintained in '/clusters/{cluster_id}/Bricks/all/{host_name}/{brick_path}/devices' in etcd as synced by gluster-integration.
3. collectd plugin uses psutil to get disk io counters.
4. io counters are available at disk level only for un-partitioned disks and partition level only for partitioned disks(which actually makes sense)

Now given 1,2, 3 and 4 above, if a brick is created from a partitioned disk, the disk stats at brick level will not be available.

Possible fixes:

- Approach 1
    * gluster-integration syncs partition details to etcd.
    * collectd plugin searches for device io stats in psutil o/p if not present uses partition details synced by gluster-integration and pushes the same to graphite.

- Approach 2(Interim - until gluster-integration adds capability of syncing partition level info):
    * collectd plugin fetches device name from etcd for every brick.
    * If the device entry not in psutil o/p(device is partitioned):
        * Iterate on each key in psutil o/p take sum of partition stats with name matching regex: 
          '%s[0-9]+' % dev_name
          Note: Partition name follows the pattern device name followed by integer numbers
          ex: vda -> vda1, vda2 etc... Cc @nnDarshan @brainfunked @shtripat @nthomas-redhat @r0h4n 
        * Use above mentioned value(sum) as particular io stat for the disk.

anmolbabu:
Closing this issue as the PR to fix this has been merged

PR #620, opened 2017-09-13
Fix partitioned disk io stat

anmolbabu:
Problem:
1. gluster-integration syncs only name of the block device corresponding to the
   brick and does not sync details of the partition of the disk from which the
   brick is carved out.
2. Collectd plugin for gathering disk stats for a brick relies on devices
   maintained in etcd at path:
   '/clusters/{cluster_id}/Bricks/all/{host_name}/{brick_path}/devices' as
   synced by gluster-integration.
3. collectd plugin uses psutil to get disk io counters.
4. io counters are available at disk level only for un-partitioned disks and
   partition level only for partitioned disks(which actually makes sense)

Approach:
1. collectd plugin fetches device name from etcd for every brick.
2. If the device entry not in psutil o/p(device is partitioned):
   a. Iterate on each key in psutil o/p take sum of partition stats with name
      matching regex: '%s[0-9]+' % dev_name
3. Use above mentioned value(sum) as particular io stat for the disk.

tendrl-bug-id: Tendrl/node-agent#619

### DIFF
--- a/tendrl/node_agent/monitoring/collectd/collectors/gluster/tendrl_gluster_brick_disk_stats.py
+++ b/tendrl/node_agent/monitoring/collectd/collectors/gluster/tendrl_gluster_brick_disk_stats.py
@@ -1,6 +1,7 @@
 import ast
 import etcd
 import psutil
+import re
 import socket
 import threading
 import time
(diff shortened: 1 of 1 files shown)


anmolbabu:
@shtripat @nnDarshan @nthomas-redhat @r0h4n Please review

anmolbabu:
Verified

Issue #628, opened 2017-09-21
Brick Disk stats collectd plugin sums all partitions for a partitioned disk stats


anmolbabu:
Brick Disk stats collectd plugin sums all partitions for a partitioned disk stats.
A brick could also be created out of 1 or more but not all partitions of a disk and summing all partitions stats to get disk stats for a brick might not be very appropriate.
Use the partitions synced by gluster-integration instead.

anmolbabu:
Closing this issue as the PR to fix this has been merged.

PR #629, opened 2017-09-21
Fix brick_disk_stats collectd plugin

anmolbabu:
Brick Disk stats collectd plugin sums all partitions for a partitioned disk stats.
A brick could also be created out of 1 or more but not all partitions of a disk
and summing all partitions stats to get disk stats for a brick might not be very
appropriate. Use the partitions synced by gluster-integration instead.

tendrl-bug-id: Tendrl/node-agent#628


anmolbabu:
To be verified.

@nnDarshan @r0h4n @nthomas-redhat @shtripat Please review

anmolbabu:
@r0h4n Yes Verified
Please merge it

Issue #638, opened 2017-10-05
Rebalance counters from get-state apply to node-level under volume and not to volume itself


anmolbabu:
Rebalance counters from get-state apply to node-level(node on which get-state is run) under volume and not to volume itself although, the key(for ex: volume1.rebalance.failures) for the same in get-state o/p seems to suggest that it applies to volume itself.

r0h4n:
Is this resolved @anmolbabu

anmolbabu:
Yes the PR to fix this on collectd side for which this issue was raised is merged. Closing this now.

PR #636, opened 2017-10-03
Add code comments

anmolbabu:
Signed-off-by: anmolbabu <anmolbudugutta@gmail.com>

PR #639, opened 2017-10-05
Push rebalance data at node level under volume and not volume level

anmolbabu:
tendrl-bug-id: Tendrl/node-agent#638


anmolbabu:
Verified

anmolbabu:
Please merge this after https://github.com/Tendrl/node-agent/pull/636
as they are rebased in that order

PR #658, opened 2017-10-26
Mark bricks of node that went down as stopped.

nnDarshan:
This patch submits a job for gluster-integration
that can mark the status of nodes that are down
as Stopped

tendrl-bug-id: Tendrl/monitoring-integration#169
Signed-off-by: Darshan N <darshan.n.2024@gmail.com>


nnDarshan:
@r0h4n @shtripat @nthomas-redhat Pls review

tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
```diff
@@ -1,20 +1,54 @@
 import etcd
 
+from tendrl.commons.objects.job import Job
+import uuid
+
+
+def update_brick_status(fqdn, integration_id):
```
r0h4n: I dont think this belongs here, please add this to https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/gluster

check
https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/ceph  (for more details)

nnDarshan: How will gluster_integration figure out about the node going down ?
Here node_agent "monitor" node already keeps track of any node going down and fires a job for "gluster_integraion" to update the status.

r0h4n: I meant to move the from here to https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/gluster/sds_sync/\__init\__.py 

It will run in the node-agent process itself, you will have to start the above sync like here https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/manager/__init__.py#L97

tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
```diff
@@ -1,20 +1,54 @@
...
+        "type": "sds"
+    }
+    Job(
+        job_id=_job_id,
+        status="new",
+        payload=_job_payload
+    ).save()
+
 
 def run():
```
r0h4n: copy this run() to https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/gluster/sds_sync/check_all_managed_brick_status.py

tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
nnDarshan: This is the only place where we get to know the point in time when node went down. If we move it to the place you suggested(in fact I tried that and during verification realized till the node comes up we will continuously fire jobs on every sync) we will end up firing jobs to mark bricks as down on every sync
Pls suggest

nnDarshan: @r0h4n Please provide your inputs

r0h4n: @nnDarshan Firstly, Copy this entire function (including node status check) to https://github.com/Tendrl/node-agent/tree/master/tendrl/integrations/gluster/sds_sync/%5C__init%5C__.py  as I had mentioned in earlier comment

And in update_brick_status(), if brick status is down, dont send the job

nnDarshan: @r0h4n If I am checking the brick status before firing the job, I can mark the status as down from here itself. why bother about the job.
What do you feel ?

r0h4n: @nnDarshan We need the job, the brick object is owned (any changes) by Gluster namespace. Please call the flow

nnDarshan:
@r0h4n Pls review. Have incorporated the suggestions and verified it

tendrl/integrations/gluster/sds_sync/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+                    # check if the bricks of this node are already
+                    # marked as down
+
+                    bricks = NS._int.client.read(
+                        "clusters/{0}/Bricks/all/{1}".format(
+                            tendrl_context.integration_id,
+                            node_context.fqdn
+                        )
+                    )
+
```
r0h4n: Only if the cluster is_managed == "yes"  then check for Brick status

https://github.com/Tendrl/gluster-integration/blob/master/tendrl/gluster_integration/sds_sync/__init__.py#L254

tendrl/node_agent/manager/__init__.py
```diff
@@ -102,6 +104,11 @@ def main():
 
     complete = threading.Event()
 
+    if "tendrl/monitor" in NS.node_context.tags:
```
r0h4n: You'll need to check

NS.tendrl_context.sds_name == "gluster"

before starting the gluster sync thread

tendrl/node_agent/manager/__init__.py
```diff
@@ -116,6 +123,9 @@ def shutdown(signum, frame):
             NS.tendrl_context.integration_id:
             NS.ceph_integrations_sync_thread.stop()
 
+        if "tendrl/monitor" in NS.node_context.tags:
```
r0h4n: You'll need to check

NS.tendrl_context.sds_name == "gluster"

before starting the gluster sync thread

tendrl/node_agent/manager/__init__.py
nnDarshan: @r0h4n Do you want this sync thread to be run on all the gluster nodes ?
I felt its enough if we run this thread only on monitor node

tendrl/integrations/gluster/sds_sync/__init__.py
nnDarshan: ack

tendrl/integrations/gluster/sds_sync/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+class GlusterIntegrtaionsSyncThread(sds_sync.StateSyncThread):
+    def run(self):
+        logger.log(
+            "debug",
+            NS.get("publisher_id", None),
+            {"message": "%s running" % self.__class__.__name__}
+        )
+        while not self._complete.is_set():
+            time.sleep(30)
```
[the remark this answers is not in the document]
nnDarshan: ack

tendrl/node_agent/manager/__init__.py
shtripat: If we have all the required details on the monitor node, it looks fine to me to run this on monitor node only.

r0h4n: ok, then we need check if the gluster_integration_sync thread has started and if not then start it.
We need to move this logic to https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/node_sync/__init__.py#L100

nnDarshan: Could you pls provide the reasoning behind this suggestion ? I see the ceph sync thread is started here: https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/manager/__init__.py#L101

Also what is the point of checking if thread is running on every node sync ?

r0h4n: - You'll not see the "tendrl/monitor" tag before the node_sync. So this code is not even going to run if you put it in manager.

- The reason to check if the thread is started/running in the node_sync is because you dont want to start a new instance of gluster_integration_thread every sync

- The ceph_integration thread start code above also needs to be moved to the node_sync, but not sure if it belongs on the monitor because it tries to do integration_id specific things. Id prefer if you remove this code all together, we dont support ceph anymore.

nnDarshan: ack

nnDarshan:
@r0h4n updated as suggested. Pls review

Issue #663, opened 2017-11-02
Need Lock for particular alert while writing in etcd


GowthamShanmugam:
Node-agent and notifier modules are reading and updating alert information in etcd under /alerting/alerts directory,
Notifier module is kept looking alerts directory 30 mins once. Lets example node-agent will update warning alert in etcd. notifier will read /alerts/be alerting and try to notify the warning alert, at same time warning alert is updated by node-agent as info alert (clearing alert) but it does not know by notifier because still notifier is busy with notifying warning alert via notification channel. When notifier updates the delivery flag in alert as true, it is actually updating alert is true in clearing alert obj, because warning alert is replaced by clearing alert. So notification for clearing alert is never pushed by the notifier.

GowthamShanmugam:
@r0h4n @shtripat i modified issue because volume alerts are pushed only from provisioning node, so we don't need any locking for that. the lock is needed only to avoid notifier and node-agent alert object update.

PR #664, opened 2017-11-02
Added lock for particular alert to avoid collision while updating 

GowthamShanmugam:
tendrl-bugid: Tendrl/node-agent#663

### DIFF
--- a/tendrl/node_agent/alert/__init__.py
+++ b/tendrl/node_agent/alert/__init__.py
@@ -20,12 +20,7 @@ class InvalidAlertSeverity(Exception):
 
 def update_alert(message):
     try:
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:

GowthamShanmugam:
Tested

tendrl/node_agent/alert/__init__.py
```diff
@@ -63,6 +65,16 @@ def update_alert(message):
                         new_alert_obj,
                         curr_alert
                     ):
+                        # Lock only if new alert matches with existing alert
+                        lock = Lock(
+                            NS._int.client,
```
r0h4n: use NS._int.wclient to lock , NS._int.client is for read-only etcd connection

Issue #688, opened 2017-11-28
Modify collectd plugin for volume heal info to use `heal info` and `heal info split-brain` options


shtripat:
The currently used command `gluster v heal <volname> statistics` is not supported. We rather should use the below commands

- gluster v heal <volname> info
- gluster v heal <volname> info split-brain

and accordingly show these values in grafana dashboard

shtripat:
@cloudbehl ^^^

nthomas-redhat:
Closing the issue as the fix is already in

PR #689, opened 2017-11-28
Changed the logic to get volume heal info

shtripat:
Earlier it was using the command `gluster v heal <volname> statistics`
But the same command is not supported. Changed the logic to use two
command instead
 - gluster v heal <volname> info
 - gluster v heal <volname> info split-brain

These details would be merged and reported as below entries in graphite
 - cluster.{int-d}.volumes.{vol-name}.brick.{brick-name}.heal_pending_cnt
 - cluster.{int-d}.volumes.{vol-name}.brick.{brick-name}.split_brain_cnt

Corresponding changes required in tendrl grafana dashboard to show these
two values instead of `heal_failed_cnt`, `healed_cnt` and `split_brain_cnt`

tendrl-bug-id: Tendrl/node-agent#688


shtripat:
@nnDarshan @anmolbabu plz review
@cloudbehl ^^^

shtripat:
tested once but need to test once with actual data.

PR #693, opened 2017-12-01
Deafult set cluster status as unhealthy if deleted due to TTL

shtripat:

### DIFF
--- a/tendrl/integrations/gluster/check_cluster_status.py
+++ b/tendrl/integrations/gluster/check_cluster_status.py
@@ -1,0 +1,32 @@
+import etcd
+
+from tendrl.commons.utils import etcd_utils
+
+
+def run():
+    try:
+        clusters = etcd_utils.read("/clusters")
+    except etcd.EtcdKeyNotFound:
+        return
+
+    # This logic only runs on node with tag `tendrl/monitor` (tendrl server)
+    # So its done checked for all the clusters and tries to set cluster
+    # status as `unhealthy` if the status field is missing (due to TTL)
+    for cluster in clusters.leaves:
+        int_id = cluster.key.split('/')[-1]
+        fetched_cluster = NS.tendrl.objects.Cluster(
+            integration_id=int_id
+        ).load()
+        try:
+            if fetched_cluster and fetched_cluster.is_managed == "yes":
+                NS._int.client.write(
+                    "/clusters/{0}/GlobalDetails/status".format(
+                        int_id
+                    ),
+                    "unhealthy",
+                    prevExist=False
+                )
+        except etcd.EtcdAlreadyExist:
+            pass
+
+    return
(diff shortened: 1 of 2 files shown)


tendrl/integrations/gluster/check_cluster_status.py
```diff
@@ -0,0 +1,16 @@
+import etcd
+
+
+def run():
+    try:
+        NS._int.client.write(
+            "/clusters/{0}/GlobalDetails/status".format(
+                NS.tendrl_context.integration_id
```
nnDarshan: This is monitor node, I guess it wont have integration id. I feel you have to loop through all the clusters and set the status

shtripat: Ack. Will do that.

shtripat:
Verified as per https://github.com/Tendrl/gluster-integration/pull/526#issuecomment-348681186

tendrl/integrations/gluster/check_cluster_status.py
```diff
@@ -0,0 +1,23 @@
+import etcd
+
+
+def run():
+    try:
+        clusters = NS._int.client.read("/clusters")
```
r0h4n: use commons.utils.etcd_util.read

r0h4n: add a comment that this is only going to be run by the "tendrl/monitor'

tendrl/integrations/gluster/check_cluster_status.py
```diff
@@ -0,0 +1,23 @@
...
+    except etcd.EtcdKeyNotFound:
+        return
+
+    for cluster in clusters.leaves:
+        int_id = cluster.key.split('/')[-1]
+        try:
+            NS._int.client.write(
```
r0h4n: do this only for is_managed == "yes" clusters

tendrl/integrations/gluster/check_cluster_status.py
shtripat: Done

shtripat: Sure

tendrl/integrations/gluster/check_cluster_status.py
shtripat: Ah ok. will change the logic and update

tendrl/integrations/gluster/check_cluster_status.py
```diff
@@ -0,0 +1,29 @@
...
+    except etcd.EtcdKeyNotFound:
+        return
+
+    # This logic only runs on node with tag `tendrl/monitor` (tendrl server)
+    # So its done checked for all the clusters and tries to set cluster
+    # status as `unhealthy` if the status field is missing (due to TTL)
+    for cluster in clusters.leaves:
+        int_id = cluster.key.split('/')[-1]
+        try:
+            if cluster.is_managed == "yes":
```
r0h4n: I am not sure how you have tested this. etcd_utils.read returns EtcdResult. you cannot do a "cluster.is_managed" directly. Thoughts?

shtripat: Updated the changes

PR #701, opened 2017-12-06
Added check for blank key or None value of heal info details

shtripat:


shtripat:
Verified this with actual setup having split brain scenario

Issue #728, opened 2018-02-26
Handle detected cluster to integration id mapping via provisioner node only


r0h4n:
Currently, all nodes ensure the map their detected cluster ID to Tendrl integration id. This needs to be centralized to the provisioner node and handled there

PR #729, opened 2018-02-26
Handle detected cluster to integration id mapping via provisioner node only

r0h4n:
tendrl-bug-id: Tendrl/node-agent#728

### DIFF
--- a/tendrl/node_agent/node_sync/sds_detect.py
+++ b/tendrl/node_agent/node_sync/sds_detect.py
@@ -44,7 +44,7 @@ def sync():
                             "%s" % sds_details['detected_cluster_id']
                         _ptag = "provisioner/%s" % \
                             NS.tendrl_context.integration_id
-                        _flow = "tendrl.flows."
+                        _flow = "tendrl.flows." \
                             "ExpandClusterWithDetectedPeers"
                         if _ptag in NS.node_context.tags:
                             dc = NS.tendrl.objects.DetectedCluster().load()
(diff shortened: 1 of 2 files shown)


tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -37,49 +38,64 @@ def sync():
...
-                            NS.tendrl_context.cluster_name = sds_details.get(
-                                'detected_cluster_name')
-                            NS.tendrl_context.sds_name = sds_details.get(
-                                'pkg_name')
-                            NS.tendrl_context.sds_version = sds_details.get(
-                                'pkg_version')
-                            NS.tendrl_context.save()
+
+                                
+                        while True:
```
shtripat: How do we make sure the above job is completed or failed? Should we wait indefinitely here?

r0h4n: Yes, the result of this job is crucial for the cluster, so we should wait indefinitely

shtripat:
Also while identifying a new peer in cluster and when we fire a import, we should raise an event I feel, that the detected node is auto imported.

r0h4n:
@shtripat That event should be triggered at the end of import job for new nodes or during ExpandCluster flow

shtripat:
I feel better we have at the end of expand cluster flow because we are sure, expand has successfully happened by that time.

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -42,39 +42,39 @@ def sync():
                         integration_index_key = \
                             "indexes/detected_cluster_id_to_integration_id/" \
                             "%s" % sds_details['detected_cluster_id']
-
-                        if "provisioner/%s" % NS.tendrl_context.integration_id \
-                            in NS.node_context.tags:
+                        _ptag = "provisioner/%s" % \
+                            NS.tendrl_context.integration_id
+                        if _ptag in NS.node_context.tags:
```
shtripat: @GowthamShanmugam is the issue mentioned around this condition resolved? Do we need some fix here?

PR #732, opened 2018-02-28
Handle detected cluster to integration id mapping via provisioner nod…

r0h4n:
tendrl-bug-id: Tendrl/node-agent#728

### DIFF
--- a/tendrl/node_agent/node_sync/sds_detect.py
+++ b/tendrl/node_agent/node_sync/sds_detect.py
@@ -43,17 +43,23 @@ def sync():
                     integration_index_key = \
                         "indexes/detected_cluster_id_to_integration_id/" \
                         "%s" % sds_details['detected_cluster_id']
+                    dc = NS.tendrl.objects.DetectedCluster().load()
+                    if dc is None or dc.detected_cluster_id is None:
+                        integration_id = str(uuid.uuid4())
+                        try:
+                            NS._int.wclient.write(
+                                integration_index_key,
+                                integration_id,
+                                prevExist=False
+                            )
+                        except etcd.EtcdAlreadyExist:
+                            pass
+
                     _ptag = "provisioner/%s" % \
                         NS.tendrl_context.integration_id
                     _flow = "tendrl.flows." \
                         "ExpandClusterWithDetectedPeers"
                     if _ptag in NS.node_context.tags:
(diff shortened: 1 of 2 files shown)


tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -35,112 +35,117 @@ def sync():
...
-                                        break
-                            else:
-                                integration_id = str(uuid.uuid4())
-                                etcd_utils.write(integration_index_key,
-                                                 integration_id)
+                try:
+                    integration_index_key = \
+                        "indexes/detected_cluster_id_to_integration_id/" \
+                        "%s" % sds_details['detected_cluster_id']
+                    _ptag = "provisioner/%s" % \
```
GowthamShanmugam: @shtripat @r0h4n my doubt is here we are checking provisioner tag but integration id is generated at line no 53 only 
https://github.com/Tendrl/node-agent/pull/732/files#diff-3dfeadf899a1fe4e98d3f57feef0b4daR53

GowthamShanmugam: oh ok when dc_id is none we are creating uuid

shtripat: Yes I think first time integration_id would be None and so provisioner/None will not in tags list

GowthamShanmugam: initially, it is None only, So here only we are creating integration_id

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -35,112 +35,123 @@ def sync():
...
+                            _job_id = str(uuid.uuid4())
+                            _job = Job(
+                                job_id=_job_id,
+                                status="new",
+                                payload=payload
+                            ).save()
+                            while True:
+                                _job = _job.load()
+                                if _job.status in ["finished",
+                                                   "failed"]:
```
shtripat: If job fails how user come to know about the same? I feel we need some kind of event logging here in case of failure. In case of success anyway flow would log an event already.

r0h4n: If a Job fails in Tendrl, there is an event created already. Did you mean we need something more than that?

shtripat: Ah thats enough I think

PR #742, opened 2018-03-13
Alert if new peers found and set cluster state if ready for import

shtripat:

### DIFF
--- a/tendrl/node_agent/node_sync/sds_detect.py
+++ b/tendrl/node_agent/node_sync/sds_detect.py
@@ -1,3 +1,4 @@
+import json
 import time
 import uuid
 
@@ -5,8 +6,8 @@ import etcd
 
 from tendrl.commons.event import Event
 from tendrl.commons.message import ExceptionMessage
-from tendrl.commons.objects.job import Job
 from tendrl.commons.utils import etcd_utils
+from tendrl.commons.utils import event_utils
 from tendrl.commons.utils import log_utils as logger
 from tendrl.node_agent.discovery.sds import manager as sds_manager
 
@@ -61,11 +62,6 @@ def sync(sync_ttl):
                         _ptag = "provisioner/%s" % \
                             NS.tendrl_context.integration_id
 
-                        _target = "tendrl/node_%s" % \
-                            NS.node_context.node_id
-                        _flow = "tendrl.flows." \
-                            "ExpandClusterWithDetectedPeers"
-
                         if _ptag in NS.node_context.tags:
                             if dc.detected_cluster_id and \
                                 dc.detected_cluster_id != sds_details.get(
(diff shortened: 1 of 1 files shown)


tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -78,32 +80,48 @@ def sync(sync_ttl):
...
+                                    msg,
+                                    "cluster_{0}".format(integration_id),
+                                    "WARNING",
+                                    integration_id=integration_id
+                                )
+                            nodes_ids = json.loads(NS._int.client.read(
+                                "indexes/tags/tendrl/integration/%s" % integration_id
+                            ).value)
+                            cluster_peers = NS._int.client.read(
+                                "/clusters/%s/Peers" % integration_id
```
r0h4n: NS.gluster.objects.Peer is not available unless tendrl-gluster-intergration is running.  Put this in a try catch and also use etcd_utils everywhere in your code

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -78,32 +80,48 @@ def sync(sync_ttl):
...
+                                event_utils.emit_event(
+                                    "cluster_status",
+                                    "expand_pending",
+                                    msg,
+                                    "cluster_{0}".format(integration_id),
+                                    "WARNING",
+                                    integration_id=integration_id
+                                )
+                            nodes_ids = json.loads(NS._int.client.read(
+                                "indexes/tags/tendrl/integration/%s" % integration_id
```
GowthamShanmugam: Here integration_id referenced before assignment

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -78,32 +80,48 @@ def sync(sync_ttl):
...
+                            ).value)
+                            cluster_peers = NS._int.client.read(
+                                "/clusters/%s/Peers" % integration_id
+                            )
+                            peer_count = 0
+                            for peer in cluster_peers.leaves:
+                                peer_count += 1
+                            if len(node_ids) == peer_count:
```
GowthamShanmugam: *nodes_ids

GowthamShanmugam:
@shtripat If any log messages which is sent by node-agent is directly passed to the logger, we are not passing then the message into socket. The logic which to parse the alert in node-agent is called by handler only. So the message not stored into alerts.

these alerts are not stored in alerts

shtripat:
Do you mean there is no way to raise alerts from node agent? This is very much required I feel.

GowthamShanmugam:
@shtripat we can do one thing if log priority is notice then we will put the alert into the node-agent socket because we can't call any node-agent file in common so the only way is for any notice log message which came from node-agent we will put that in socket

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -78,32 +80,50 @@ def sync(sync_ttl):
...
+                                "/clusters/%s/Peers" % integration_id
+                            )
+                            peer_count = 0
+                            for peer in cluster_peers.leaves:
+                                peer_count += 1
+                            _cluster = NS.tendrl.objects.Cluster(
+                                integration_id=integration_id
+                            ).load()
+                            if len(nodes_ids) == peer_count and \
+                                _cluster.status != "expand_pending":
```
GowthamShanmugam: gluster status is always "expand_pending", because when we import cluster it always len(nodes_ids) == peer_count
Once this blcok is executed cluster status is marked as expand_pending in etcd, so when new node came for expand then no clearing alert raised

shtripat: Ah yes. So when new peers detected we need to set the cluster status as `new_peers_detected` say and when no of peers equals the no of nodes in indexes and cluster status is `new_peer_detected` then only we need to set it to `expand_pending`. Sent the PR with changes below.

PR #745, opened 2018-03-15
Corrected logic for raising clearing alert for new peers

shtripat:
The peers list in tendrl central store would take sometime to reflect
and initially the no od peers would be equal to no of managed nodes
of cluster. So using this condition would cause issue. Rather we should
check the no of managed nodes of cluster against actual peers list from
underlying cluster and once they are same raise the clearing alert.

tendrl-bug-id: Tendrl/commons#805

### DIFF
--- a/tendrl/node_agent/node_sync/sds_detect.py
+++ b/tendrl/node_agent/node_sync/sds_detect.py
@@ -1,8 +1,9 @@
+import etcd
 import json
+import subprocess
 import time
 import uuid
 
-import etcd
 
 from tendrl.commons.event import Event
 from tendrl.commons.message import ExceptionMessage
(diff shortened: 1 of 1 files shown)


shtripat:
Verified the changes and works as expected.
When a new nodes is peer probed in cluster an alert is raised saying new nodes detected. This alerts remain till tendrl-node-agent is started on the new node. Once tendrl-node-agent is started on the new node, this alert is cleared with another one saying cluster is ready for expand now.

shtripat:
@r0h4n @GowthamShanmugam @nthomas-redhat plz review

GowthamShanmugam:
LGTM

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -94,42 +95,50 @@ def sync(sync_ttl):
...
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE
                                 )
-                                # Set the cluster status accordingly
-                                _cluster.status = 'expand_pending'
-                                _cluster.save()
+                                out, err = cmd.communicate()
+                                if err or out is None or \
+                                    "Connection failed" in out:
+                                    pass  # set the no of peers as zero
```
r0h4n: I think you wanted to use "continue" or "return"?

shtripat: Actually its fine if we pass as in that case `peers` would be empty list and the condition at L#121 will not pass and no clearing alert raised. We dont want to stop the logic from L#143 from execution in this case. Correct?

r0h4n: L#113 will raise exception if out is None, handle that and then the pass makes sense

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -94,42 +95,50 @@ def sync(sync_ttl):
...
+                                lines = out.split('\n')[1:]
+                                for line in lines:
+                                    if line != '':
```
r0h4n: line.strip()

tendrl/node_agent/node_sync/sds_detect.py
shtripat: Ah yes. will add a check

tendrl/node_agent/node_sync/sds_detect.py
shtripat: ok

Issue #746, opened 2018-03-16
Storing cluster node alert count separately 


GowthamShanmugam:
Currently, cluster_node_alert count is stored in node_alert_count obj, the problem in this approach is node_alert_count object does not have integration_id and object is stored using integration_id in tendrl_context, the problem this approach will work only for storage nodes. All utilization alerts are stored in the server node, the server node unable to increment the cluster_node_alert_count.

GowthamShanmugam:
This issue is fixed

PR #747, opened 2018-03-16
Storing cluster node alert count separately

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#746


GowthamShanmugam:
@r0h4n @shtripat please review

GowthamShanmugam:
Tendrl/monitoring-integration/issues/374

GowthamShanmugam:
Currently, cluser_node_alert count stored using integration_id from tendrl_context. Problem is server node can't increment alert counts for utilization alerts. Passing integration id to  node_alert_count object is also wrong, to solve this problem cluster_node_alert_count obj is stored separately.

GowthamShanmugam:
Verified

PR #750, opened 2018-03-22
Added logic to consider gluster_get_ver_command to get sds version

shtripat:
The command could be different to find out the sds version details
downstream, so made this configurable and accordingly read and set
the sds version information.

### DIFF
--- a/tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
+++ b/tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
@@ -49,13 +49,13 @@ class DiscoverGlusterStorageSystem(DiscoverSDSPlugin):
         ret_val['peers'] = gfs_peers
 
         cmd = subprocess.Popen(
-            "gluster --version",
+            "rpm -qa | grep glusterfs-server",
             shell=True,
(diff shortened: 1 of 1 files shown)


shtripat:
NEED TO VERIFY

tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
```diff
@@ -43,8 +43,23 @@ def discover_storage_system(self):
         ret_val['detected_cluster_id'] = cluster_id
         ret_val['detected_cluster_name'] = "gluster-%s" % cluster_id
 
+        gluster_get_ver_cmd = NS.config.data.get(
+            "gluster_get_ver_command",
```
r0h4n: config?

NS.compiled_definitions.get_parsed_defs()['namespace.tendrl'].get("gluster_get_ver_command")

shtripat: Ah my bad it should be from definitions only. Anyway its not verified yet and discussed this might not be a workable solution and we need more discussion based on BZ# 1559507.

shtripat:
Closing this as will be handled differently

Issue #753, opened 2018-03-28
If one time node alert calculation is goes wrong then it always shows wrong 


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed, i am closing this

PR #754, opened 2018-03-28
Calculating and updtating alert_count in every node sync

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#753

### DIFF
--- a/tendrl/node_agent/node_sync/__init__.py
+++ b/tendrl/node_agent/node_sync/__init__.py
@@ -154,7 +154,8 @@ class NodeAgentSyncThread(sds_sync.StateSyncThread):
                 sync_cluster_contexts_thread.start()
                 sync_cluster_contexts_thread.join()
             # Update node alert count
-            update_node_alert_count()
+            if not NS.tendrl.objects.NodeAlertCounters().exists():
+                update_node_alert_count()
             time.sleep(_sleep)
         logger.log(
             "info",
--- a/tendrl/node_agent/node_sync/__init__.py
+++ b/tendrl/node_agent/node_sync/__init__.py
@@ -2,13 +2,10 @@ import threading
 import time
 
 from etcd import EtcdException
-from etcd import EtcdKeyNotFound
 
 from tendrl.commons.event import Event
 from tendrl.commons.message import ExceptionMessage
-from tendrl.commons.objects.node_alert_counters import NodeAlertCounters
 from tendrl.commons import sds_sync
-from tendrl.commons.utils import etcd_utils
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @shirshendu @r0h4n please review

GowthamShanmugam:
Tendrl/monitoring-integration/issues/373

GowthamShanmugam:
Each node will update read its own alerts only and it will update its own alerts count only

tendrl/node_agent/node_sync/__init__.py
```diff
@@ -164,10 +153,39 @@ def run(self):
                 sync_cluster_contexts_thread.daemon = True
                 sync_cluster_contexts_thread.start()
                 sync_cluster_contexts_thread.join()
-
+            # Update node alert count
+            update_node_alert_count()
```
shtripat: Do you really want to do this every sync as an overwrite, or re-calculate only if node is deleted due to TTL? @r0h4n comments?

r0h4n: Should be done only if NS.tendrl.objects.NodeAlertCounters.exists() is False

GowthamShanmugam: i will change this

GowthamShanmugam:
@shtripat @r0h4n changes done

tendrl/node_agent/node_sync/__init__.py
```diff
@@ -164,10 +153,40 @@ def run(self):
                 sync_cluster_contexts_thread.daemon = True
                 sync_cluster_contexts_thread.start()
                 sync_cluster_contexts_thread.join()
-
+            # Update node alert count
+            if not NS.tendrl.objects.NodeAlertCounters().exists():
```
r0h4n: You should be increment the alert count if NS.tendrl.objects.NodeAlertCounters().exists():  

right?

GowthamShanmugam: @r0h4n When node details are deleted by ttl or initially if alert count object is not present so at that time only it will save alert count object in etcd.

GowthamShanmugam: What i feel is let it calculate alert count in every sync because it calculating alert count only for that node only, it is not reading all alerts. So whatever happens alert count won't goes wrong

@shtripat @r0h4n

GowthamShanmugam: just a suggestion

PR #757, opened 2018-04-02
Support gluster peer fqdn/ip

r0h4n:
tendrl-bug-id: Tendrl/commons#852

### DIFF
--- a/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
+++ b/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
@@ -21,17 +21,6 @@ def _parse_heal_info_stats(tree):
         brick_host = brick_name.split(":")[0]
         brick_path = brick_name.split(":")[1]
 
-        # If brick host is returned as an IP conver to FQDN
-        try:
-            from dns import resolver
-            from dns import reversename
-            from IPy import IP
-            IP(brick_host)
-            addr = reversename.from_address(brick_host)
-            brick_host = str(resolver.query(addr, "PTR")[0])[:-1]
-        except ValueError:
-            pass
-
         no_of_entries = 0
         try:
             no_of_entries = int(brick.find("numberOfEntries").text)
--- a/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
+++ b/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
@@ -1,4 +1,5 @@
 import collectd
+import socket
 import time
 import traceback
 # import threading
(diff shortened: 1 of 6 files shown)


r0h4n:
@shtripat @GowthamShanmugam please review, WIP

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -39,6 +39,42 @@ def sync(sync_ttl):
...
+                        NS.tendrl_context.integration_id
+                    _node_ids = etcd_utils.read(integration_id_index_key).value
+                    _node_ids = json.loads(_node_ids)
+                    peer_node_id = ""
+                    for _node_id in _node_ids:
+                        if _node_id != NS.node_context.node_id:
+                            peer_node_id = _node_id
+                            break
+
+                    if peer_node_id:
```
shtripat: Is it like in one peer we maintain one other's peer's uuid? Not getting clear to me what purpose it serves. Can you please add some code level comments and explain?

r0h4n: Every node agent will save the gluster pool list. And to identify the hostname for current node we query any other node in cluster to find out

shtripat: Where is this pool list saved?

r0h4n: Check this PR and this snippet

https://github.com/Tendrl/commons/pull/891
https://github.com/Tendrl/commons/pull/891/files#diff-1a5320aa0fd1eea178e6ed255509c9c4R1192

shtripat: Ah yes. Get that point.

GowthamShanmugam: So all nodes have gluster peer object and to avoid localhost name we are reading gluster peer object for the current node from another node. This is my understanding.

GowthamShanmugam: looks good to me

tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
```diff
@@ -21,17 +21,6 @@ def _parse_heal_info_stats(tree):
...
         brick_path = brick_name.split(":")[1]
 
-        # If brick host is returned as an IP conver to FQDN
-        try:
-            from dns import resolver
-            from dns import reversename
-            from IPy import IP
-            IP(brick_host)
-            addr = reversename.from_address(brick_host)
-            brick_host = str(resolver.query(addr, "PTR")[0])[:-1]
```
GowthamShanmugam: all details are stored in graphite using fqdn which we are passing, so while storing brick details also we have to replace brick hostname using fqdn and store it. otherwise, if the brick hostname is different from fqdn then it won't stored in a proper hierarchy.

GowthamShanmugam:
In collected wherever we are storing brick details we need to replace brick hostname with fqdn which we are passing. We can't store brick details as it is. Because it won't comes under proper hierarchy.

tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
shtripat: Ah yes agree. I missed somehow. this reverse naming would be required.

Issue #769, opened 2018-04-11
Collectd is not pushing brick utilization details when peer hostname and brick hostname not match


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed

PR #770, opened 2018-04-11
Fixing collectd plugin to push brick details when brick hostname and peer hostname not match

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#769

Signed-off-by: Gowthamshanmugasundaram <gshanmug@redhat.com>


GowthamShanmugam:

Issue #778, opened 2018-04-14
Check for existence of configuration file for underlying storage cluster version details


shtripat:
If the file exists read values from there else default with rpm version details

shtripat:
Closing as required PRs are merged

PR #779, opened 2018-04-14
Changed the logic to figure out the sds version and name

shtripat:
tendrl-bug-id: Tendrl/node-agent#778

### DIFF
--- a/tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
+++ b/tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
@@ -1,4 +1,6 @@
 import hashlib
+import json
+import os
 import subprocess
 
 from tendrl.commons.utils import log_utils as logger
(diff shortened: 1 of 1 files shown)


tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
```diff
@@ -48,23 +50,46 @@ def discover_storage_system(self):
...
-        )
-        out, err = cmd.communicate()
-        if err and 'command not found' in err:
-            logger.log(
-                "debug",
-                NS.publisher_id,
-                {"message": "gluster not installed on host"}
+        # Check if the file /etc/tendrl/redhat-gluster-storage.json
+        # if the file exists, read the version details from this
+        if os.path.exists('/etc/tendrl/redhat-gluster-storage.json'):
```
r0h4n: which rpm is going to provide this file? If it is not a Tendrl rpm then it should not be inside `/etc/tendrl`

tendrl/node_agent/discovery/sds/plugins/discover_gluster_sds_details.py
```diff
@@ -48,23 +50,46 @@ def discover_storage_system(self):
...
+            except IOError:
+                logger.log(
+                    "debug",
+                    NS.publisher_id,
+                    {"message": "Failed to get version details"}
+                )
+                return ret_val
+        else:
+            cmd = subprocess.Popen(
+                "rpm -qa | grep glusterfs-server",
```
r0h4n: add comment with sample

shtripat: Ack

PR #781, opened 2018-04-17
Removed ttl for node context

shtripat:
Because FQDN and IP gets removed after sometime due to TTL

### DIFF
--- a/tendrl/node_agent/node_sync/__init__.py
+++ b/tendrl/node_agent/node_sync/__init__.py
@@ -59,7 +59,7 @@ class NodeAgentSyncThread(sds_sync.StateSyncThread):
             current_tags += ["tendrl/node_%s" % NS.node_context.node_id]
             NS.node_context.tags = list(set(current_tags))
             NS.node_context.status = "UP"
-            NS.node_context.save(ttl=_sync_ttl)
+            NS.node_context.save()
             NS.tendrl_context = NS.tendrl_context.load()
 
             sync_service_and_index_thread = threading.Thread(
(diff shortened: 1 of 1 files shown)


tendrl/node_agent/node_sync/__init__.py
```diff
@@ -59,7 +59,7 @@ def run(self):
             current_tags += ["tendrl/node_%s" % NS.node_context.node_id]
             NS.node_context.tags = list(set(current_tags))
             NS.node_context.status = "UP"
-            NS.node_context.save(ttl=_sync_ttl)
+            NS.node_context.save()
```
r0h4n: ttl is required for NodeContext.status

shtripat: After sometime I have seen FQDN and IP getting empty for cluster nodes, and setting these values again happens only if cluster node context's is_managed says `no`.
So once these fields go blank, there is no way they could be populated back again. And because of this brick sync in gluster integration starts failing (as it needs equality check with current node's fqdn with brick's node).

shtripat: Also by any chance if node agent gets re-started on node, as manager does NS.node_context.save() this will set fqdn and ip as blank again, and as the node is already managed as part of cluster, its not going to be populated back again.

shtripat: If we dont want to remove the TTL, in that case can we remove the if condition at https://github.com/Tendrl/node-agent/blob/master/tendrl/node_agent/node_sync/sds_detect.py#L47 to make sure fdqn and ip is populated again and again? It saves the peers of node before node is managed by cluster. This we can keep doing post import as well. This remains a way because we decide FQDN and IP of a node based on its peers providing this information. (BTW this peer telling about fqdn and ip of a node doesnt work for a single node cluster ;) )

PR #785, opened 2018-04-18
Corrected checking of sds_name

shtripat:
tendrl-bug-id: Tendrl/node-agent#778


tendrl/node_agent/alert/utils.py
```diff
@@ -131,7 +131,7 @@ def update_alert_count(alert, existing_alert=None):
             sds_name = find_sds_name(
                 alert.tags['integration_id']
             )
-        if sds_name == constants.GLUSTER:
+        if sds_name in [constants.GLUSTER, "RHGS"]:
```
nthomas-redhat: Can you put RHGS into constants and access from there

Issue #804, opened 2018-05-10
import cluster flow timeout during executing a flow for setup alias for cluster


GowthamShanmugam:
Setup alias job is always in a new state, it is not taken by a monitoring-integration, so import cluster flow is timeout

shtripat:
I feel this is not the scenario we had when it failed. In a croner case when services like tendrl-node-agent and tendrl-monitoring-integration were started back to back very fast and so the server node was not having the tag `tendrl/integration/monitoring`. To tackle this situation we need to make sure at every service sync of the node-agent if tendrl-monitoring-integration found running, it should try to add tag `tendrl/integration/monitoring`.

Does this sound good enough??

GowthamShanmugam:
This issue is fixed, i am closing this

GowthamShanmugam:
This issue is fixed

PR #805, opened 2018-05-10
Flag for monitoring-integration is assigned by node-agent

GowthamShanmugam:
flag is assigned in common but it not assigned for server

tendrl-bug-id : Tendrl/node-agent#804
bugzilla: 1570048


tendrl/node_agent/node_sync/services_and_index_sync.py
```diff
@@ -15,6 +15,7 @@
     "etcd",
     "tendrl-api",
     "tendrl-gluster-integration",
+    "tendrl-monitoring-integration",
```
GowthamShanmugam: flag is added in common definition file to assign flag but without checking service it won't add

Issue #820, opened 2018-05-29
Tendrl Dashboard showing wrong info and alert details.


anmolsachan:
Tendrl doesn't reflect that all gluster nodes are down.
Sometimes Shows :Alerts: Node None is DOWN

PR #821, opened 2018-05-29
Updated logic for alert notifications and node status watch

anmolsachan:
tendrl-bug-id: #820 
Bugzilla-id: 1519201, 1583264, 1580385
Signed-off-by: Anmol Sachan <anmol13694@gmail.com>

### DIFF
--- a/tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
+++ b/tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
@@ -1,11 +1,18 @@
-def run():
-    nodes = NS.tendrl.objects.Node().load_all()
-    if nodes is None:
-        return
+import etcd
 
-    for node in nodes:
-        _node_context = NS.tendrl.objects.NodeContext(
-            node_id=node.node_id
-        ).load()
-        _node_context.watch_attrs()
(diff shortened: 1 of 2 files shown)


tendrl/node_agent/node_sync/check_all_managed_nodes_status.py
```diff
@@ -1,11 +1,18 @@
...
-        return
+import etcd
+
+from tendrl.commons.utils import etcd_utils
 
-    for node in nodes:
-        _node_context = NS.tendrl.objects.NodeContext(
-            node_id=node.node_id
-        ).load()
-        _node_context.watch_attrs()
```
GowthamShanmugam: When node_agent starts it does not have fqdn in node_context, after SDS detection and only it will have fqdn, the problem with this approach is it creates a watchable thread with node_context object which does not all the details. Whenever it calls on change function it will override the node_context detail with none values. That is why sometimes we saw fqdn None problem in gluster-integration sync.

anmolsachan:
@r0h4n Need to verify with commons changes, will notify after verifying.

anmolsachan:
@r0h4n Verified.

PR #832, opened 2018-06-28
Optimize collectd metrics push logic(WIP)

GowthamShanmugam:


shtripat:
How this would work for the scenario where few plugins get executed only from provisioner node and few run from all the storage nodes. Still we might see not all data coming in grafana dashboards at the same time. Right?

GowthamShanmugam:
It can't be same time it will differ based on sleep time

Issue #850, opened 2018-09-20
Somtimes node-agent message socket file "message.sock" is missing


GowthamShanmugam:
When node-agent service is started socket file "message.sock" was created under /var/run/tendrl. Using socket file only node-agent message socket server listen and accept the connections. Sometimes when we restart node-agent socket file is not created and node-agent message handler thread is failing.

I have seen this scenario a few times after installation done from tendrl-ansible in a fresh machine.

All other components in a node are kept raising the following exception:

Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: Traceback (most recent call last):
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: File "/usr/lib/python2.7/site-packages/tendrl/commons/event.py", line 38, in _write
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: self.sock.connect(self.socket_path)
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: File "/usr/lib64/python2.7/socket.py", line 224, in meth
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: return getattr(self._sock,name)(*args)
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: error: [Errno 2] No such file or directory
Sep 16 03:42:02 GlusterWebAdmin tendrl-monitoring-integration: Unable to pass the message into socket.{"integration_id": null, "publisher": "monitoring_integration", "job_id": null, "timestamp": "2018-09-16T03:42:02.051258+00:00", "caller": {"function": "load_definition", "line_no": 50, "filename": "/usr/lib/python2.7/site-packages/tendrl/commons/objects/__init__.py"}, "payload": {"message": "Load definitions (.yml) for namespace.tendrl.objects.Job"}, "priority": "debug", "parent_id": null, "node_id": "a6ed6301-260d-4a4d-8a35-62f980122ee1", "flow_id": null, "message_id": "a178660d-7333-43d0-a7ef-aace791fbf30"}

GowthamShanmugam:
this issue is fixed

PR #851, opened 2018-09-20
Fixing node-agent socket problem after temporary network issue

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#850
bugzilla: 1630344

### DIFF
--- a/tendrl/node_agent/message/handler.py
+++ b/tendrl/node_agent/message/handler.py
@@ -90,15 +90,19 @@ class MessageHandler(threading.Thread):
         # http://0pointer.de/blog/projects/systemd.html (search "file
         # descriptor 3")
         try:
-            socket_fd = 3
-            self.sock = socket.fromfd(socket_fd, socket.AF_UNIX,
-                                      socket.SOCK_STREAM)
-            self.sock.listen(50)
-            return self.sock
+            # systemd socket should exist before listen
+            if os.path.exists(MESSAGE_SOCK_PATH):
+                socket_fd = 3
+                self.sock = socket.fromfd(socket_fd, socket.AF_UNIX,
+                                          socket.SOCK_STREAM)
+                self.sock.listen(50)
+                return self.sock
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @nthomas-redhat @anmolsachan please review

tendrl/node_agent/message/handler.py
```diff
@@ -90,15 +90,19 @@ def bind_unix_listener(self):
         # http://0pointer.de/blog/projects/systemd.html (search "file
         # descriptor 3")
         try:
-            socket_fd = 3
-            self.sock = socket.fromfd(socket_fd, socket.AF_UNIX,
-                                      socket.SOCK_STREAM)
-            self.sock.listen(50)
-            return self.sock
+            # systemd socket should exist before listen
+            if os.path.exists(MESSAGE_SOCK_PATH):
```
shtripat: Who creates this file? Sometimes we have seen the socket file doesnt exist, even if tendrl-node-agent is running. How we tackle that scenario?

GowthamShanmugam: message.socket file is created by node-agent.socket systemd file, node-agent systemd file is invoking that. The reason which we are not seen message socket is sometime node-agent start won't invoking node-agent.socket systemd file but node-agent restart will invoke this always.

When connection issue happening with etcd then node-agent will keep trying to connect after few minutes node-agent will start again(this scenario I saw in a customer machine) so node-agent start is not invoking node-agent.socket systemd so message.sock file is not created. 

As per this patch if message.sock file does not present then below code https://github.com/Tendrl/node-agent/pull/851/files#diff-3eae70b9dbe53a810f8c236acc9baa6dR103 will tack care of creating. So message.sock file will always present.

shtripat: Ok. Looks fine.

GowthamShanmugam:
@shtripat i agree

GowthamShanmugam:
@nthomas-redhat please review and merge this PR

PR #856, opened 2018-11-02
Each collectd plugins are executed more than once in each sync

GowthamShanmugam:


tendrl/node_agent/monitoring/collectd/collectors/gluster/tendrl_gluster.py
```diff
@@ -272,7 +272,10 @@ def r_callback():
     """
     init()
     read_callback("/low_weight", "low_weight")
+    # clear low_weight plugin import
+    destroy()
```
shtripat: If we dont have this `destroy` call, is it like line#277 would execute the plugins again?

GowthamShanmugam: It will execute but it will execute low_weight  + heavy_weight plugins, destroy will remove low_weight plugins from a list for  line  #277 will execute only heavy_weight plugins.

shtripat: In that case shouldn't we just remove the line#274 ? `read_callback("/low_weight", "low_weight")`

GowthamShanmugam: But we need to execute low weight plugin also, if we remove line #274  then how we can invoke that, the actual execution is low weight plugin executed and the heavyweight plugins are also executed.

shtripat:
pep8 fixes

shtripat:
Fix the pep8 issues reported here

GowthamShanmugam:
@shtripat pep8 issue fixed for same file in #857

Issue #867, opened 2019-01-18
Tendrl log files needs log rotation


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed

PR #868, opened 2019-01-18
Logrotate for tendrl log files

GowthamShanmugam:
bugzilla: 1654183
tendrl-bug-id: Tendrl/node-agent#867

### DIFF
--- a/tendrl-node-agent.spec
+++ b/tendrl-node-agent.spec
@@ -54,6 +54,7 @@ install -Dm 0644 etc/tendrl/node-agent/logging.yaml.syslog.sample $RPM_BUILD_ROO
 install -Dm 644 etc/tendrl/node-agent/*.sample $RPM_BUILD_ROOT%{_datadir}/tendrl/node-agent/
 mkdir -p $RPM_BUILD_ROOT/%{_sysconfdir}/rsyslog.d
 install -Dm 644 etc/rsyslog.d/tendrl-node-agent.conf $RPM_BUILD_ROOT/%{_sysconfdir}/rsyslog.d/tendrl-node-agent.conf
+install -Dm 0644 etc/logrotate.d/tendrl-node-agent_logrotate.conf $RPM_BUILD_ROOT%{_sysconfdir}/logrotate.d/tendrl-node-agent_logrotate.conf
 cp -a tendrl/node_agent/monitoring/collectd/collectors/* $RPM_BUILD_ROOT%{_libdir}/collectd/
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@dahorak @mbukatov @fbalak

GowthamShanmugam:
@shtripat @nthomas-redhat @TimothyAsirJeyasing please review

GowthamShanmugam:
i tested using a command:
   logrotate  -v  /etc/logrotate.d/tendrl_logrotate.conf
it will fire logrotation manually for that configuration file

tendrl-node-agent.spec
```diff
@@ -90,6 +91,7 @@ py.test -v tendrl/node-agent/tests || :
 %{_unitdir}/tendrl-node-agent.service
 %{_unitdir}/tendrl-node-agent.socket
 %config(noreplace) %{_sysconfdir}/rsyslog.d/tendrl-node-agent.conf
+%config(noreplace) %{_sysconfdir}/logrotate.d/tendrl_logrotate.conf
 
 %changelog
 * Sat Jul 14 2018 Rohan Kanade <rkanade@redhat.com> - 1.6.3-9
```
shtripat: Add a changelog entry as well.
Otherwise look fine

tendrl-node-agent.spec
```diff
@@ -90,8 +91,12 @@ py.test -v tendrl/node-agent/tests || :
 %{_unitdir}/tendrl-node-agent.service
 %{_unitdir}/tendrl-node-agent.socket
 %config(noreplace) %{_sysconfdir}/rsyslog.d/tendrl-node-agent.conf
+%{_sysconfdir}/logrotate.d/tendrl-node-agent_logrotate.conf
```
GowthamShanmugam: i am not giving %config(noreplace) because while removing node-agent it should be removed I feel, otherwise next time while installing node-agent we will see two log rotation files.

TimothyAsirJeyasing:
LGTM, Kindly test it once with local build and install

tendrl-node-agent.spec
shtripat: Ack. I feel tendrl should have control of when to rotate its logs files.

GowthamShanmugam:
I created local build and verified:
the command which I used is:
           logrotate -vf /etc/logrotate.d/tendrl_logrotate.conf

tendrl-node-agent.spec
shtripat: @GowthamShanmugam I feel logrotate daemon doesnt consider files with .rpmsave so we can have %config(noreplace) %{_sysconfdir}/logrotate.d/tendrl-node-agent_logrotate.conf

https://forums.fedoraforum.org/archive/index.php/t-298877.html

PR #869, opened 2019-01-19
After logrotate logging is not happening

GowthamShanmugam:
bugzilla: 1654183
tendrl-bug-id: Tendrl/node-agent#867

### DIFF
--- a/etc/logrotate.d/tendrl-node-agent_logrotate.conf
+++ b/etc/logrotate.d/tendrl-node-agent_logrotate.conf
@@ -6,5 +6,5 @@
     missingok
     compress
     notifempty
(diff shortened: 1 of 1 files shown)


etc/logrotate.d/tendrl-node-agent_logrotate.conf
```diff
@@ -6,5 +6,5 @@
     missingok
     compress
     notifempty
-    create 0664 root root
+    copytruncate
```
GowthamShanmugam: instead of creating a new file,  it will keep the old file and continue logging. So create has no effect here. https://linux.die.net/man/8/logrotate

Issue #870, opened 2019-01-25
Tendrl stops monitoring when glusterd is down in provisioner node 


GowthamShanmugam:
I have created 3 nodes gluster cluster. After the import flow, I have manually brought down the glusterd service in provisioner node. After a few minutes, volume details are disappeared from the tendrl-UI.

In grafana, volumes related panels are also affected. 

When glusterd is down, Provisioner node is unable to execute gluster commands. So it stops collecting the gluster cluster topology and monitoring metrics. All Other nodes are unable to claim the provisioner tag. 

Request ETCD find the provisioner node:
   1. Read provisioner key to find node_id
        "indexes/tags/provisioner"
   2. Read fqdn key fro node_context:
         "nodes/{node_id}/NodeContext/fqdn"

GowthamShanmugam:
For now, I am closing this issue.

PR #871, opened 2019-01-25
Claim provisioner tag only by a node which peer status is connected

GowthamShanmugam:
bugzilla: 1669453
tendrl-bug-id: Tendrl/node-agent#870

### DIFF
--- a/tendrl/node_agent/node_sync/services_and_index_sync.py
+++ b/tendrl/node_agent/node_sync/services_and_index_sync.py
@@ -155,8 +155,28 @@ def sync(sync_ttl=None):
                         continue
                 if NS.node_context.node_id in _node_ids:
                     if sync_ttl and len(_node_ids) == 1:
-                        etcd_utils.refresh(index_key,
-                                           sync_ttl + 50)
+                        refresh_tag = True
+                        if "provisioner" in tag:
+                            _cluster = _cluster.load()
+                            if _cluster.is_managed == "yes":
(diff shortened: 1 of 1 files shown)


tendrl/node_agent/node_sync/services_and_index_sync.py
```diff
@@ -155,8 +155,28 @@ def sync(sync_ttl=None):
...
+                            _cluster = _cluster.load()
+                            if _cluster.is_managed == "yes":
+                                # Allow to claim provisioner tag only by a
+                                # node which have peer status is connected
+                                # in a managed cluster
+                                peers = NS.tendrl.objects.GlusterPeer(
+                                    node_id=NS.node_context.node_id
+                                ).load_all()
+                                for peer in peers:
+                                    if peer.hostname == 'localhost':
```
GowthamShanmugam: own nodes peer hostname is localhost only

GowthamShanmugam: this check only for the managed cluster, after a cluster, is managed only the node which has a peer status is "connected" can claim the provisioner.

GowthamShanmugam: in unmanage cluster if any one glusterd is down anyway import will fail, so we don't need to check this

GowthamShanmugam:
@shtripat @nthomas-redhat @dahorak @anmolsachan please revirew

Issue #859, opened 2018-11-08
Collectd heavy weight plugins are executed by non-provisioner nodes also


GowthamShanmugam:

GowthamShanmugam:
This issue is fixed

PR #874, opened 2019-02-18
Modified collectd logic to read node_id from etcd instead of reading from conf file

GowthamShanmugam:
bugzilla: 1647910
tendrl-bug-id: Tendrl/node-agent#859

### DIFF
--- a/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
+++ b/tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
@@ -61,8 +61,11 @@ def get_volume_heal_info_split_brain_stats(vol, integration_id, etcd_client):
     except (
         AttributeError,
         KeyError,
-        ValueError
+        ValueError,
+        ElementTree.ParseError
     ):
+        # For heal info command timeout and older version of glusterd
+        # ElementTree will raise parser error
         collectd.error(
             'Failed to collect volume heal info split-brain. Error %s' % (
                 traceback.format_exc()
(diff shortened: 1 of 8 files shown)


tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
```diff
@@ -61,7 +62,8 @@ def get_volume_heal_info_split_brain_stats(vol, integration_id, etcd_client):
     except (
         AttributeError,
         KeyError,
-        ValueError
+        ValueError,
+        ElementTree.ParseError
```
GowthamShanmugam: I saw some unhandled exceptions, so i added ElementTree.ParseError

GowthamShanmugam:
@shtripat @nthomas-redhat @dahorak please review, There is 2 different commits in this PR, one for an actual fix and one for upstream pep8 fix

tendrl/node_agent/monitoring/collectd/collectors/gluster/heavy_weight/tendrl_gluster_heal_info.py
shtripat: In what scenario this exception can happen? I feel you should add some comment as well to explain that.

GowthamShanmugam: I will comment, This will happen when heal info command gives None. During gluster heal info command  timeout or lower version of gluster without nolog command support this error will come

GowthamShanmugam:
@shtripat Changes done

dahorak:
What component is configuring the `node_id` in the collectd configuration files? And is it used anywhere else? Otherwise LGTM.

GowthamShanmugam:
@dahorak we are not going to read node_id form configuration file, it is not used anywhere

PR #877, opened 2019-04-19
Un-manage should not fail when any one or few storage nodes are down

GowthamShanmugam:
tendrl-bug-id: Tendrl/node-agent#876
bugzilla: 1645221

### DIFF
--- a/tendrl/node_agent/node_sync/sds_detect.py
+++ b/tendrl/node_agent/node_sync/sds_detect.py
@@ -106,7 +106,6 @@ def sync(sync_ttl, node_status_ttl):
                             if dc.detected_cluster_id and \
                                 dc.detected_cluster_id != sds_details.get(
                                     'detected_cluster_id'):
-
                                 # Gluster peer list has changed
                                 integration_id = \
                                     NS.tendrl_context.integration_id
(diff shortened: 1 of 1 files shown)


tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -106,34 +106,36 @@ def sync(sync_ttl, node_status_ttl):
...
                                     integration_id
                                 )
-                                # Set the cluster status as new peer detected
                                 _cluster = NS.tendrl.objects.Cluster(
                                     integration_id=integration_id
                                 ).load()
-                                _cluster.status = "new_peers_detected"
-                                _cluster.save()
-                                # Raise an alert regarding the same
-                                msg = "New peers identified in cluster: %s. " \
```
shtripat: So you want to say only if cluster managed then cluster expansion required? makes sense..

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -212,6 +214,32 @@ def sync(sync_ttl, node_status_ttl):
                     detected_cluster_tag = "detected_cluster/%s" % \
                                            sds_details[
                                                'detected_cluster_id']
+                    # Detected cluster id will change when new node
+                    # added into peer list and when peer detach happens,
+                    # Node_context should not maintain multiple DC ids
+                    old_dc_id = "detected_cluster/%s" % dc.detected_cluster_id
```
shtripat: I see that `old_dc_id` you get from `detected_cluster/{int_id}` and at line#214 `detected_cluster_tag` also as `detected_cluster/{sds_detail's detected_cluster_id}`? Do you mean if nodes removed / added from CLI the value of `sds_details['detected_cluster_id']` would differ and so no match between `old_dc_id` and `detected_cluster_tag`?

tendrl/node_agent/node_sync/sds_detect.py
```diff
@@ -212,6 +214,32 @@ def sync(sync_ttl, node_status_ttl):
...
+                        indexes_keys = []
+                        indexes_keys.append(
+                            "indexes/detected_cluster_id_to_integration_id"
+                            "/%s" % dc.detected_cluster_id
+                        )
+                        indexes_keys.append(
+                            "indexes/tags/detected_cluster/%s" %
+                            dc.detected_cluster_id
+                        )
+                        for indexes_key in indexes_keys:
```
shtripat: What is this code for ? Add some comments to explain.

shtripat:
Comments not replied or taken care. Please check.

tendrl/node_agent/node_sync/sds_detect.py
GowthamShanmugam: yes, let's take an example if we have two nodes detected. if user peer probe another node before managing a cluster then it will display two clusters as detected. Because multiple detected cluster_id will present in ETCD.

tendrl/node_agent/node_sync/sds_detect.py
GowthamShanmugam: yes, an old-detected cluster is no longer when we added a new node.  So sds_details['detected_cluster_id'] should be replaced by old-id. No need to maintain both the detected cluster_id ETCD.

tendrl/node_agent/node_sync/sds_detect.py
GowthamShanmugam: I have already added here: 
https://github.com/Tendrl/node-agent/pull/877/files/c7ba6aec5f40c4a8a10d4404f586f47f04b9b10e#diff-3dfeadf899a1fe4e98d3f57feef0b4daR217

https://github.com/Tendrl/node-agent/pull/877/files/c7ba6aec5f40c4a8a10d4404f586f47f04b9b10e#diff-3dfeadf899a1fe4e98d3f57feef0b4daR224

